PR #1, opened 2018-07-26
Set up test suite

zuren:
Removes the default test suite Rails installs (minitest) and sets up RSpec (much better!)

Bear in mind most of the files are generated by rspec installation (`rails generate rspec:install`)

Particularly interested in review of the documentation + make commands. I use this method a lot but haven't documented it before.


Makefile
```diff
@@ -8,3 +8,9 @@ serve:
 bundle:
 	docker-compose run --rm app bundle
 	docker-compose build
+
+test:
+	docker-compose run --rm app rspec
+
+shell:
+	docker-compose exec app /bin/bash
```
StevenLeighton21: Doesn't this need `-i` to not release the process? I assume it works, I'm just surprised

zuren: `docker exec` would - it'd have to be `docker exec -it {container_id} /bin/bash`.

`docker-compose exec` is interactive by default though.

StevenLeighton21: Didn't know that, cool!

README.md
```diff
@@ -2,14 +2,32 @@
...
+To run tests:
+
+```sh
+$ make test
 ```
+
+If you're TDDing code, it can sometimes be faster to boot up the app container once, then run tests within it. That way you don't have to start the docker container every time you run tests:
+
+```sh
 $ make serve
```
StevenLeighton21: It's not abundantly clear _where_ to run `rspec`

PR #5, opened 2018-08-07
Determine priority scores using SQL queries

zuren:
- Migrate TenancyPrioritiser to use SQL instead of hashes from the Old Income API.
- Pretty performant over a local VPN, so expect it to be better in staging and production.


lib/hackney/income/tenancy_prioritiser/universal_housing_criteria.rb
```diff
@@ -0,0 +1,158 @@
...
+        def balance
+          attributes.fetch(:current_balance).to_f
+        end
+
+        def days_in_arrears
+          day_difference(Date.today, attributes.fetch(:pre_arrears_date))
+        end
+
+        def days_since_last_payment
+          return nil if attributes.fetch(:last_payment_date).nil?
```
zuren: Had to add this guard for situations where `last_payment_date` was null. Suspect this might just be test cases.

lib/hackney/income/tenancy_prioritiser/universal_housing_criteria.rb
```diff
@@ -0,0 +1,158 @@
...
+              @BreachedAgreementsCount as breached_agreements_count,
+              @NospsInLastYear as nosps_in_last_year,
+              @NospsInLastMonth as nosps_in_last_month,
+              @Payment1Value as payment_1_value,
+              @Payment1Date as payment_1_date,
+              @Payment2Value as payment_2_value,
+              @Payment2Date as payment_2_date,
+              @Payment3Value as payment_3_value,
+              @Payment3Date as payment_3_date;
+          SQL
```
zuren: Need to guard against SQL injection here.

lib/hackney/income/tenancy_prioritiser/universal_housing_criteria.rb
```diff
@@ -0,0 +1,158 @@
...
+        end
+
+        # FIXME: implementation needs confirming, will return to later
+        def broken_court_order?
+          false
+        end
+
+        private
+
+        ACTIVE_ARREARS_AGREEMENT_STATUS = '200'.freeze
```
StevenLeighton21: Suggest we add these to a constants file, if only because there will be more to add as we go and I'm sure it'll come in handy to have them live in one place.

lib/hackney/income/tenancy_prioritiser/universal_housing_criteria.rb
```diff
@@ -0,0 +1,158 @@
+module Hackney
+  module Income
+    class TenancyPrioritiser
+      class UniversalHousingCriteria
+        def self.for_tenancy(universal_housing_client, tenancy_ref)
```
StevenLeighton21: This is really impressive. I believe I understand what it's doing, but don't feel confident I could offer critique on improving it or catch any _gotchas_.
What a power move 💯

spec/hackney/income/tenancy_prioritiser/score_spec.rb
```diff
@@ -0,0 +1,261 @@
+require 'rails_helper'
+
+describe Hackney::Income::TenancyPrioritiser::Score do
```
StevenLeighton21: Well written specs though 🥇

lib/hackney/income/tenancy_prioritiser/universal_housing_criteria.rb
zuren: Thanks!

PR #6, opened 2018-08-07
Expose tenancies with priorities through use case

zuren:
~**Must merge https://github.com/LBHackney-IT/lbh-income-api/pull/5 first, then rebase this before reviewing.**~

- Syncs tenancy priorities locally.
- Gathers tenancy details from the Tenancy API and merges in local priorities.
- API endpoint to come in the next PR.


lib/hackney/income/dangerous_view_my_cases.rb
```diff
@@ -0,0 +1,36 @@
...
+      end
+
+      def execute(tenancy_refs)
+        full_tenancies = @tenancy_api_gateway.get_tenancies_by_refs(tenancy_refs)
+        stored_tenancies = @stored_tenancies_gateway.get_tenancies_by_refs(tenancy_refs)
+
+        full_tenancies.map do |tenancy, index|
+          stored_tenancy = stored_tenancies.find { |t| t.fetch(:tenancy_ref) == tenancy.fetch(:ref) }
+
+          {
```
StevenLeighton21: would it be better to match the format of the other API output here? ie. `latest_action: { code: }` is `{ latest_action_code }` in the output from the TenancyAPI

lib/hackney/income/universal_housing_tenancies_gateway.rb
```diff
@@ -0,0 +1,21 @@
+module Hackney
+  module Income
+    class UniversalHousingTenanciesGateway
+      def tenancies_in_arrears
+        tenancy_refs = []
+
+        query = universal_housing_client.execute('SELECT tag_ref FROM tenagree WHERE cur_bal > 0')
+        query.each { |record| tenancy_refs << record.fetch('tag_ref').strip }
```
StevenLeighton21: Assume future PR to filter this to `active` `secure` tenancies?

zuren: Indeed ^

lib/hackney/income/dangerous_view_my_cases.rb
zuren: It isn't as far as I can see? At least in the list endpoint. Mirrored the Tenancy API output exactly here, just appended priority band and score.

spec/hackney/income/tenancy_api_gateway_spec.rb
```diff
@@ -0,0 +1,129 @@
...
+  context 'when retrieving tenancies' do
+    subject { gateway.get_tenancies_by_refs(refs) }
+
+    context 'and using a different host' do
+      let(:gateway) { described_class.new(host: 'https://other.com') }
+      let(:refs) { %w(123) }
+
+      before do
+        stub_request(:get, "https://other.com/api/v1/tenancies?tenancy_refs%5B%5D=123")
+          .to_return(body: { 'tenancies' => [example_tenancy] }.to_json)
```
StevenLeighton21: Example tenancy doesn't match actual JSON from this API - not sure if this is a cause for concern

spec/hackney/income/tenancy_api_gateway_spec.rb
```diff
@@ -0,0 +1,129 @@
...
+          primary_contact: nil
+        }])
+      end
+    end
+
+    context 'and two tenancy refs are given' do
+      let(:refs) { %w(000015/01 000017/01) }
+
+      before do
+        stub_request(:get, "https://example.com/api/v1/tenancies?tenancy_refs%5B%5D=000017/01&tenancy_refs%5B%5D=000015/01")
```
StevenLeighton21: When stubbing requests for single tenancies, I've had to encode `/` characters. I'm sure that this works fine because they're params, but should we encode anyway for safety?

lib/hackney/income/dangerous_view_my_cases.rb
StevenLeighton21: Yeah I had assumed it would use the detailed output but that doesn't make a lot of sense, so disregard my comment

PR #12, opened 2018-08-14
Add production deploy

lukemorton:
We add a new build step to circle config to add production deployments of lbh-income-api.

We also include another change to configure `ecs-deploy` timeouts via ENV:

`ecs-deploy` will timeout is a deploy does not complete within 90 seconds by default. It can take longer than this as we not only have to drain traffic from old deploys but also wait for ELB to deregister and register new applications. See here for more information: https://github.com/fabfuel/ecs-deploy/issues/14#issuecomment-277424439

We allow the timeout to be set via ENV in Circle as we've had issues of deploys taking more than 10 minutes to fully drain old tasks. Putting this in ENV will allow us to configure dynamically without changing the .circle/config.yml in future.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -102,9 +102,9 @@ workflows:
       - staging_release:
           requires:
             - check
-          # filters:
-          #   branches:
-          #     only: master
+          filters:
+            branches:
+              only: master
       - permit_production_release:
           type: approval
           requires:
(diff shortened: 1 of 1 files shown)


StevenLeighton21:
Are you sure it's definitely time related? Is the production task set to 0% minimum available? That was what caused staging to fail to release previously (it needs to be 0% as there's only one instance, so if it's set to any higher AWS won't remove the old container, and the new will fail to build)

PR #19, opened 2018-08-28
Add daily sync cases job

zuren:
- Syncs cases using the DangerousSyncCases use case.
- Runs daily at 3am. May need to amend this time, but fine for now I think.
- Schedules itself on completion, regardless of success.
- Fails without retrying at the moment, to avoid retries rolling over into normal daily operating hours.
- Logs errors.

### DIFF
--- a/app/jobs/hackney/income/jobs/sync_cases_job.rb
+++ b/app/jobs/hackney/income/jobs/sync_cases_job.rb
@@ -1,0 +1,32 @@
+module Hackney
+  module Income
+    module Jobs
+      class SyncCasesJob < ApplicationJob
+        queue_as :default
+
+        after_perform do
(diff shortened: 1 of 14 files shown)


app/jobs/hackney/income/jobs/sync_cases_job.rb
```diff
@@ -0,0 +1,29 @@
...
+
+        def perform
+          sync_cases.execute
+        rescue => e
+          Rails.logger.error("Caught error: #{e}")
+        ensure
+          self.class.enqueue_next
+        end
+
+        def sync_cases
```
StevenLeighton21: Should probably use the use case in Hackney::Income::UseCaseFactory?

zuren: Nice one 👍

PR #21, opened 2018-08-29
Save priority factors

StevenLeighton21:
Adds a set of score contributions to the tenancy model, to be persisted along with score and band.

Saved directly as the numerical contribution to the score to be provided as a breakdown, so we can say 'having an active NOSP contributed x of y to the score' etc. 
This means by default, the score contribution of 100.00 balance would be stored as +120.00 score contribution, which should help clarify how the different score values are reached and aid tweaking score generation through feedback > adjust weighting > feedback cycle.

### DIFF
--- a/db/migrate/20180829140413_add_priority_factors_to_tenancies.rb
+++ b/db/migrate/20180829140413_add_priority_factors_to_tenancies.rb
@@ -12,11 +12,11 @@ class AddPriorityFactorsToTenancies < ActiveRecord::Migration[5.1]
     add_column :tenancies, :active_nosp_contribution, :decimal
 
     add_column :tenancies, :balance, :decimal
-    add_column :tenancies, :days_in_arrears, :decimal
-    add_column :tenancies, :days_since_last_payment, :decimal
+    add_column :tenancies, :days_in_arrears, :integer
+    add_column :tenancies, :days_since_last_payment, :integer
     add_column :tenancies, :payment_amount_delta, :decimal
-    add_column :tenancies, :payment_date_delta, :decimal
-    add_column :tenancies, :number_of_broken_agreements, :decimal
+    add_column :tenancies, :payment_date_delta, :integer
+    add_column :tenancies, :number_of_broken_agreements, :integer
     add_column :tenancies, :active_agreement, :boolean
     add_column :tenancies, :broken_court_order, :boolean
     add_column :tenancies, :nosp_served, :boolean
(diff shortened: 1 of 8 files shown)


db/migrate/20180829140413_add_priority_factors_to_tenancies.rb
```diff
@@ -0,0 +1,14 @@
...
+    add_column :tenancies, :days_since_last_payment_contribution, :decimal
+    add_column :tenancies, :payment_amount_delta_contribution, :decimal
+    add_column :tenancies, :payment_date_delta_contribution, :decimal
+    add_column :tenancies, :number_of_broken_agreements_contribution, :decimal
+    add_column :tenancies, :active_agreement_contribution, :decimal
+    add_column :tenancies, :broken_court_order_contribution, :decimal
+    add_column :tenancies, :nosp_served_contribution, :decimal
+    add_column :tenancies, :active_nosp_contribution, :decimal
+  end
+end
```
zuren: Wonder if these should be in a separate table?

i.e.

`tenancies`: id, ref, priority_score, priority_band
`tenancy_priority_criteria`: tenancy_id, balance_contribution, days_in_arrears_contribution, etc...

lib/hackney/income/universal_housing_prioritisation_gateway.rb
```diff
@@ -7,7 +7,7 @@ def priorities_for_tenancy(tenancy_ref)
         weightings = Hackney::Income::TenancyPrioritiser::PriorityWeightings.new
         prioritiser = Hackney::Income::TenancyPrioritiser.new(criteria: criteria, weightings: weightings)
 
-        { priority_score: prioritiser.priority_score, priority_band: prioritiser.priority_band }
+        { priority_score: prioritiser.priority_score, priority_band: prioritiser.priority_band, criteria: criteria, weightings: weightings }
```
zuren: Feel it might be preferable to return a hash of criteria and weightings, rather than their objects? Not convinced, but wonder if every instance of a prioritisation gateway would necessarily use objects with their interfaces? Maybe they should?

zuren:
I actually _really_ like the idea of storing the contribution of a particular field, but think the user will want to see the value as well as it's contribution. Otherwise they still have to reverse engineer the values from the 'contribution' factor?

i.e. they'll have to think "days in arrears contributes highly - they must have been in arrears a while" when I think ideally they'd be presented with "Days In Arrears: 30 days (+60 score)"

Can we persist and serve both?

lib/hackney/income/universal_housing_prioritisation_gateway.rb
StevenLeighton21: What's the likelihood of a different gateway having a different prioritisation implementation vs a different data source? And even a new datasource with different data requirements will probably just use a new criteria, weighting etc.

I agree with the notion, but I like sticking to objects where they exist and am probably sacrificing something for brevity in the implementation.

db/migrate/20180829140413_add_priority_factors_to_tenancies.rb
StevenLeighton21: Potentially, but another table keyed on the same ID is probably overkill when the table represents our _picture_ of a tenancy. I'd expect any further info we choose to cache about a tenancy could go in here also, which makes serving the _picture_ to an API consumer simpler IMO. Open to discussion on this as I can see the reasoning for tightly scoped data storage!

PR #23, opened 2018-08-30
Restrict tenancies used to beta patches when generating priority scores

zuren:
For our Beta team, we're going to start by only generating priority scores for tenancies in Beta patches. This should help us avoid performance issues making our sync job take too long in production, and reduce initial load on the database for a first test.

- Configures which patches are restricted via env vars.
- Will need to update env vars on staging and production before releasing.


madetech-jeffpinkham:
I'm happy, we optimised the query a bit, looks good 👍

PR #28, opened 2018-09-05
Run migrations on deployment

zuren:
- [x] Set environment variables on CircleCI.
- [ ] Test staging release migrates successfully.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -59,6 +59,10 @@ jobs:
             docker push $ECR_IMAGE_URL:latest
             docker push $ECR_IMAGE_URL:staging
       - run:
+          name: Run migrations
+          command: |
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -58,6 +58,10 @@ jobs:
             docker push $ECR_IMAGE_URL:$CIRCLE_SHA1
             docker push $ECR_IMAGE_URL:latest
             docker push $ECR_IMAGE_URL:staging
+      - run:
```
StevenLeighton21: Do we need to `migrate RAILS_ENV=test`?

zuren: Nah, we won't run test on staging/production. The test db is already migrated earlier in the workflow - https://github.com/LBHackney-IT/lbh-income-api/pull/28/files#diff-1d37e48f9ceff6d8030570cd36286a61L21

PR #32, opened 2018-09-06
Fix concurrent tasks not running on ECS

zuren:
Runs a separate worker container for each environment on ECS, instead of running jobs and the application in the same container.

A) Fixes a bug where both _aren't_ running concurrently at the moment.
B) Means both can be stopped gracefully when ECS redeploys.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,4 +14,4 @@ RUN bundle check || bundle install
 COPY . /app
 EXPOSE 3000
 
-CMD ["sh", "-c", "rake jobs:work ; rails s"]
+CMD rails s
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -14,4 +14,4 @@ RUN bundle check || bundle install
 COPY . /app
 EXPOSE 3000
 
-CMD ["sh", "-c", "rake jobs:work ; rails s"]
+CMD rails s
```
StevenLeighton21: is this an intentional reversion or a merge mistake?

zuren: Intentional reversion. Removing the failed attempt at this task, as it'll now be handled in a separate container.

PR #40, opened 2018-09-13
Endpoint to find or create user

StevenLeighton21:
Endpoint to create users upon login.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,4 +1,5 @@
 Rails.application.routes.draw do
   get '/api/v1/my-cases', to: 'my_cases#index'
   get '/api/v1/sync-cases', to: 'my_cases#sync'
+  post '/api/v1/find-or-create-user', to: 'users#create'
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,5 +1,5 @@
 Rails.application.routes.draw do
   get '/api/v1/my-cases', to: 'my_cases#index'
   get '/api/v1/sync-cases', to: 'my_cases#sync'
-  post '/api/v1/find-or-create-user', to: 'users#create'
+  post '/api/v1/find-or-create', to: 'users#create'
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,5 +1,5 @@
 Rails.application.routes.draw do
   get '/api/v1/my-cases', to: 'my_cases#index'
   get '/api/v1/sync-cases', to: 'my_cases#sync'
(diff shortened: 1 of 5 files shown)


spec/controllers/users_controller_spec.rb
```diff
@@ -0,0 +1,127 @@
...
+        provider_permissions: params.fetch(:provider_permissions)
+      ).and_return(
+        provider_uid: params.fetch(:provider_uid),
+        provider: params.fetch(:provider),
+        name: params.fetch(:name),
+        email: params.fetch(:email),
+        first_name: params.fetch(:first_name),
+        last_name: params.fetch(:last_name),
+        provider_permissions: params.fetch(:provider_permissions)
+      )
```
zuren: This is incorrectly stubbed - the return value comes directly from the SqlUsersGateway (and shouldn't really) which looks like this:

```rb
{
  id: user.id,
  name: user.name,
  email: user.email,
  first_name: user.first_name,
  last_name: user.last_name,
  provider_permissions: user.provider_permissions
}
```

StevenLeighton21: I had thought it might be reasonable to actually check the contents of the model. Prefer to mock?

spec/controllers/users_controller_spec.rb
```diff
@@ -0,0 +1,127 @@
...
+
+      get :create, params: params
+    end
+
+    it 'should return the response as json' do
+      get :create, params: params
+
+      expect(response.body).to eq(
+        {
+          id: 1,
```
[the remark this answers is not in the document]
zuren: Should be whatever the Use Case is stubbed to respond with.

spec/controllers/users_controller_spec.rb
zuren: Means our controller test is tied to which gateway we're using in production now, which would cause headaches if we wanted to swap it out.

StevenLeighton21: Yeah I was thinking feature in my head, will update. Cheers!

config/routes.rb
```diff
@@ -1,4 +1,5 @@
 Rails.application.routes.draw do
   get '/api/v1/my-cases', to: 'my_cases#index'
   get '/api/v1/sync-cases', to: 'my_cases#sync'
+  post '/api/v1/find-or-create', to: 'users#create'
```
zuren: Meant /api/v1/**users**/find-or-create

StevenLeighton21: Well, if we're gonna be _sensible_

PR #41, opened 2018-09-17
Ignore terminated tenancies when running the priority sync

zuren:


zuren:
The `active` column doesn't seem correct to me. There's a `present` boolean column, which `terminated` appears to be the inverse of, which seems accurate. In this selection, the tenancies which have end of tenancy dates in the past are "terminated" but still "active".

The "terminated" ones appear to line up with EOTs in the past.

image

PR #44, opened 2018-09-18
Assign on sync

StevenLeighton21:
Add use case that attempts to assign cases to the next available user, if the case is not already assigned.

Add the use case to the factory, call it for each synced tenancy.

### DIFF
--- a/spec/lib/hackney/income/assign_tenancy_to_user_spec.rb
+++ b/spec/lib/hackney/income/assign_tenancy_to_user_spec.rb
@@ -1,0 +1,45 @@
+require 'rails_helper'
+
+describe Hackney::Income::AssignTenancyToUser do
+  let!(:user1) { Hackney::Income::Models::User.create(name: Faker::Name.name) }
+  let!(:user2) { Hackney::Income::Models::User.create(name: Faker::Name.name) }
+  let!(:assigned_tenancy) { create_assigned_tenancy_model(band: 'green', user: user2) }
+  let!(:unassigned_tenancy) { create_assigned_tenancy_model(band: 'green', user: nil) }
+
+  let(:gateway) { double('UserAssignmentGateway') }
+  subject { described_class.new(user_assignment_gateway: gateway)}
+
+  before do
+    allow(gateway).to receive(:assign_to_next_available_user).with(tenancy: unassigned_tenancy).and_return(user1.id)
+  end
+
+  context 'when trying to assign a tenancy already assigned' do
+    it 'should not assign the tenancy' do
+      expect(gateway).to_not receive(:assign_to_next_available_user)
+
+      expect(subject.assign(tenancy: assigned_tenancy)).to eq(user2.id)
+    end
+  end
+
+  context 'when trying to assign a new tenancy' do
+    it 'should pass the tenancy to the assignment gateway and return the assigned user id' do
+      expect(gateway).to receive(:assign_to_next_available_user).with(
+        tenancy: unassigned_tenancy
+      )
+
+      expect(subject.assign(tenancy: unassigned_tenancy)).to eq(user1.id)
+    end
+  end
+
+  def create_assigned_tenancy_model(band:, user:)
+    tenancy = Hackney::Income::Models::Tenancy.new.tap do |t|
+      t.tenancy_ref = Faker::Lorem.characters(5)
+      t.priority_band = band
+      t.priority_score = Faker::Lorem.characters(5)
+      t.assigned_user = user
+    end
+
+    tenancy.save
+    tenancy
+  end
+end
--- a/spec/lib/hackney/income/assign_tenancy_to_user_spec.rb
+++ b/spec/lib/hackney/income/assign_tenancy_to_user_spec.rb
@@ -1,8 +1,8 @@
 require 'rails_helper'
 
 describe Hackney::Income::AssignTenancyToUser do
-  let!(:user1) { Hackney::Income::Models::User.create(name: Faker::Name.name) }
-  let!(:user2) { Hackney::Income::Models::User.create(name: Faker::Name.name) }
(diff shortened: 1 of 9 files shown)


spec/lib/hackney/income/assign_tenancy_to_user_spec.rb
```diff
@@ -0,0 +1,45 @@
+require 'rails_helper'
+
+describe Hackney::Income::AssignTenancyToUser do
+  let!(:user1) { Hackney::Income::Models::User.create(name: Faker::Name.name) }
+  let!(:user2) { Hackney::Income::Models::User.create(name: Faker::Name.name) }
```
zuren: Recommend using `create!` and `update!` whenever possible in tests, as it'll raise an exception if it doesn't work. Low risk as we're not using any ActiveModel validation on models, but worth considering.

As it is, if the creation doesn't work for some reason, it won't persist but won't let us know.

spec/lib/hackney/income/dangerous_sync_cases_spec.rb
```diff
@@ -4,12 +4,14 @@
   let(:uh_tenancies_gateway) { double(tenancies_in_arrears: []) }
   let(:stored_tenancies_gateway) { double(store_tenancy: nil) }
   let(:prioritisation_gateway) { PrioritisationGatewayDouble.new }
+  let(:assign_tenancy_to_user) { AssignTenancyToUserDouble.new }
```
zuren: Defining and using `AssignTenancyToUserDouble` is probably overkill, as there's no custom behaviour within it.

Could use `double(assign: nil)`, which returns an object that responds to `assign` with any args and returns `nil`.

spec/lib/hackney/income/sql_tenancy_case_gateway_spec.rb
```diff
@@ -0,0 +1,158 @@
...
+
+  def create_assigned_tenancy_model(band:, user:)
+    tenancy = Hackney::Income::Models::Tenancy.new.tap do |t|
+      t.tenancy_ref = Faker::Lorem.characters(5)
+      t.priority_band = band
+      t.priority_score = Faker::Lorem.characters(5)
+      t.assigned_user = user
+    end
+
+    tenancy.save
```
zuren: This might be more readable/idiomatic as:

```rb
Hackney::Income::Models::Tenancy.create!(
  tenancy_ref: Faker::Lorem.characters(5),
  priority_band: band,
  priority_score: Faker::Lorem.characters(5),
  assigned_user: user
)
```

spec/lib/hackney/income/sql_tenancy_case_gateway_spec.rb
```diff
@@ -0,0 +1,158 @@
...
+          expect(subject.assign_to_next_available_user(tenancy: unassigned_red)).to eq(user2.id)
+          expect(unassigned_red.assigned_user).to eq(user2)
+        end
+      end
+
+      it 'should not assign if the band cannot be matched' do
+        expect(subject.assign_to_next_available_user(tenancy: unassigned_case)).to be_nil
+        expect(unassigned_case.assigned_user).to be_nil
+      end
+    end
```
zuren: I totally understand the logic involved here, but my eyes are glazing over a little looking at the test cases. Wondering if there's a more human readable way to write them, to make sure we're not missing anything.

StevenLeighton21: I agree, I think most of that comes from the setup. There may be a nice way to handle the arrange, even if it's just putting it in a function and naming things a bit nicer. I would say that I'm happy with the test descriptions outlining the logic.

zuren: Could just be me? Maybe something like:

```rb
3.times { assign_case(:green, user_1) }
2.times { assign_case(:green, user_2) }

expect_next_case_assignment_to_be(:green, user_2)
```

If it makes sense to you and you're confident, just ignore me.

zuren: The test descriptions are good, yeah. Think it's the massive amount of setup in `let`s at the top, and then the mix of logic being tested in the `it` blocks.

StevenLeighton21: Your snipped probably looks nicer, bear with me

spec/lib/hackney/income/sql_tenancy_case_gateway_spec.rb
```diff
@@ -0,0 +1,155 @@
...
+        end
+
+        it 'should behave the same way for each band' do
+          2.times { create_assigned_tenancy_model(band: 'red', user: user1) }
+          1.times { create_assigned_tenancy_model(band: 'red', user: user2) }
+
+          expect(subject.assign_to_next_available_user(tenancy: unassigned_red)).to eq(user2.id)
+          expect(unassigned_red.assigned_user).to eq(user2)
+        end
+      end
```
zuren: How about assignment when no cases have been assigned yet?

spec/lib/hackney/income/sql_tenancy_case_gateway_spec.rb
zuren: Looks good 👍

spec/lib/hackney/income/sql_tenancy_case_gateway_spec.rb
StevenLeighton21: covered by the case when the users have equal numbers of cases already.

zuren: The logic should be the same, yeah, but are we confident the SQL will work the same way/not error?

StevenLeighton21: I'll add a test case around it 👍

PR #45, opened 2018-09-18
Sync each case as a separate background job

zuren:
- The job running all cases in band is taking way too long. Giving it more time would be a poor fix, as four hours should be more than enough, and it would risk overlapping business hours.
- This updates the DangerousSyncCases use case to schedule a background job for each tenancy in arrears, so multiple workers can pick them up concurrently, rather than attempting to sync them all sequentially in the same job.
- Have been working on optimising the SQL separately, but conscious most of the bottleneck here is poor UH performance (and lack of parallelism.)

**Todo**
- [ ] Increase memory and worker count on ECS to accomodate.

### DIFF
--- a/spec/lib/hackney/income/sync_case_priority_spec.rb
+++ b/spec/lib/hackney/income/sync_case_priority_spec.rb
@@ -1,0 +1,80 @@
+require_relative '../../../../lib/hackney/income/tenancy_prioritiser/stub_criteria'
+require_relative '../../../../lib/hackney/income/tenancy_prioritiser/priority_weightings'
+require_relative '../../../../lib/hackney/income/sync_case_priority'
+
+describe Hackney::Income::SyncCasePriority do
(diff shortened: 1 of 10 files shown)


spec/lib/hackney/income/sync_case_priority_spec.rb
```diff
@@ -0,0 +1,80 @@
+require_relative '../../../../lib/hackney/income/tenancy_prioritiser/stub_criteria'
```
StevenLeighton21: Interests of quickness or? Shouldn't these just be available?

zuren: Nah, we could use `require 'rails_helper'` to load Rails and the whole app, or use `require_relative` and a path to just load that ruby file without anything else.

Helps TDDing when running one spec at a time. When running the suite it won't make a difference.

PR #46, opened 2018-09-18
Sync cases as individual background jobs nightly (with assignment)

zuren:
- The job running all cases in band is taking way too long. Giving it more time would be a poor fix, as four hours should be more than enough, and it would risk overlapping business hours.
- This updates the DangerousSyncCases use case to schedule a background job for each tenancy in arrears, so multiple workers can pick them up concurrently, rather than attempting to sync them all sequentially in the same job.
- Have been working on optimising the SQL separately, but conscious most of the bottleneck here is poor UH performance (and lack of parallelism.)

**Todo**

- [x] Increase memory and worker count on ECS to accomodate.

### DIFF
--- a/app/jobs/hackney/income/jobs/sync_case_priority_job.rb
+++ b/app/jobs/hackney/income/jobs/sync_case_priority_job.rb
@@ -1,0 +1,13 @@
+module Hackney
+  module Income
+    module Jobs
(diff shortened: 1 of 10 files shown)


zuren:
Previous PR was approved by @StevenLeighton21.

PR #49, opened 2018-09-27
Add user role column and exclude non-CC users from assignment

StevenLeighton21:
What:
Add a role column to users and exclude users who do not have a role '1:credit_controller' from case assignment.

Why:
Currently assignment assigns cases evenly across all users, while some users in the system are devs/managers/legal case workers who will not need case assignment.

Future:
The income API needs some sort of management interface, and this interface will probably need to be able to modify the roles of users along with other management tasks like re-assigning cases manually.

### DIFF
--- a/app/models/hackney/income/models/user.rb
+++ b/app/models/hackney/income/models/user.rb
@@ -3,6 +3,8 @@ module Hackney
     module Models
       class User < ApplicationRecord
         has_many :tenancies, foreign_key: :assigned_user_id
+
+        enum role: [ :developer, :credit_controller, :legal_case_worker, :manager ]
       end
     end
   end
--- a/app/models/hackney/income/models/user.rb
+++ b/app/models/hackney/income/models/user.rb
@@ -4,7 +4,7 @@ module Hackney
       class User < ApplicationRecord
         has_many :tenancies, foreign_key: :assigned_user_id
 
-        enum role: [ :developer, :credit_controller, :legal_case_worker, :manager ]
+        enum role: [:base_user, :credit_controller, :legal_case_worker, :manager, :developer]
       end
     end
   end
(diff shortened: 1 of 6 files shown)


app/models/hackney/income/models/user.rb
```diff
@@ -3,6 +3,8 @@ module Income
     module Models
       class User < ApplicationRecord
         has_many :tenancies, foreign_key: :assigned_user_id
+
+        enum role: [ :developer, :credit_controller, :legal_case_worker, :manager ]
```
zuren: White space is off here (no need to pad arrays)

db/migrate/20180927140212_add_role_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddRoleToUsers < ActiveRecord::Migration[5.1]
+  def change
+    add_column :users, :role, :integer, default: 0
+  end
```
zuren: Should we default to developer here? What if we give devs extra capabilities? NULL is probably safer IMO.

spec/lib/hackney/income/assign_tenancy_to_user_spec.rb
```diff
@@ -1,8 +1,8 @@
 require 'rails_helper'
 
 describe Hackney::Income::AssignTenancyToUser do
-  let!(:user1) { Hackney::Income::Models::User.create!(name: Faker::Name.name) }
-  let!(:user2) { Hackney::Income::Models::User.create!(name: Faker::Name.name) }
+  let!(:user1) { Hackney::Income::Models::User.create!(name: Faker::Name.name, role: 1) }
```
zuren: Should be able to provide a symbol instead of an int here - would help readability.

app/models/hackney/income/models/user.rb
StevenLeighton21: https://api.rubyonrails.org/v5.1/classes/ActiveRecord/Enum.html

They did it in the AR docs :/ cheeky!

db/migrate/20180927140212_add_role_to_users.rb
StevenLeighton21: I was under the impression you'd have to default to something, and I'd argue that if that's the worry then 0 should be something else, like `basic_user` or something.

PR #50, opened 2018-10-05
POC to update the stack level on the container

mrosel:
This is an attempt to fix `stack level too deep` when running the sql updates. The provided ruby `./bin/stack_test.rb`  should not throw an error if the env setup is successful.

TBD: if 5mb is enough

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,8 @@ ARG RAILS_ENV=development
 FROM ruby:2.5.1
 WORKDIR /app
 
+ENV RUBY_THREAD_VM_STACK_SIZE=5000000
+
 RUN wget ftp://ftp.freetds.org/pub/freetds/stable/freetds-1.00.27.tar.gz && \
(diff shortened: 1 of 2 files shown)


cob16:
Just out of curiosity is there anything we can do in our code to reduce the stack depth or is this happening in 3rd party code?

mrosel:
> Just out of curiosity is there anything we can do in our code to reduce the stack depth or is this happening in 3rd party code?

Yes, actually that's a good question because this is technically a work around. Ideally, you would favor iteration over recursion - however, this isn't the simplest SQL that I've seen and I believe this is the culprit - https://github.com/LBHackney-IT/lbh-income-api/blob/master/lib/hackney/income/tenancy_prioritiser/universal_housing_criteria.rb#L32

StevenLeighton21:
I think it's worth a try - it's entirely possible that there is a genuine issue causing those errors not just a heavy transaction list. I have also isolated an issue with the assignment that is unrelated, but this is still worth doing IMO.

PR #51, opened 2018-10-08
Resolve issue in case assignment

StevenLeighton21:
The previous assignment would work fine for the first priority batch.

Then, when assigning the next cases, the `where` clause could not evaluate to true, so they all got assigned to the first available person.

Added a test around this, and resolved by mapping a list of `{user.id, count(assigned_tenancies.id) where tenancy_bad = band}` and selecting the `min_by &:count` of this list to determine the next available user.


lib/hackney/income/sql_tenancy_case_gateway.rb
```diff
@@ -21,20 +21,19 @@ def assigned_tenancies(assignee_id:)
       def assign_to_next_available_user(tenancy:)
         return nil if !['red', 'amber', 'green'].include?(tenancy.priority_band)
 
-        tenancy.assigned_user = next_available_user_for(band: tenancy.priority_band) || Hackney::Income::Models::User.all.first
```
madetech-jeffpinkham: Was this the line that was selecting the first user and assigning all cases to them?

madetech-jeffpinkham: Ok, I don't understand what the min_by function is doing? Can you explain it to me please? It appears to be taking a count and an id and returns the lowest value int id from the list?

lib/hackney/income/sql_tenancy_case_gateway.rb
```diff
@@ -21,20 +21,21 @@ def assigned_tenancies(assignee_id:)
       def assign_to_next_available_user(tenancy:)
         return nil if !['red', 'amber', 'green'].include?(tenancy.priority_band)
```
madetech-jeffpinkham: Suggest creating a const for the priority bands?

lib/hackney/income/sql_tenancy_case_gateway.rb
```diff
@@ -21,20 +21,21 @@ def assigned_tenancies(assignee_id:)
...
       private
 
       def next_available_user_for(band:)
-        Hackney::Income::Models::User
-          .left_joins(:tenancies)
-          .group('users.id')
-          .where('users.role = 1 AND tenancies.id IS NULL OR tenancies.priority_band = ?', band)
-          .order('COUNT(tenancies.id) ASC')
-          .first
+        counts = Hackney::Income::Models::User.where(role: :credit_controller).all.map do |u|
```
mrosel: Would prefer to separate this out into it's own method as we're now swapping around types and converting from ActiveRecord, to an Array of Hashes, back to ActiveRecord. 

Better yet would be to keep it as a `User` and scope it out. `User.with_tenancy_counts`

lib/hackney/income/sql_tenancy_case_gateway.rb
StevenLeighton21: It returns the id for the user who has the lowest count :)

In effect, it maps how many 'reds' each user has, and then gives back the user whose count was lowest, as the next eligible assignee.

lib/hackney/income/sql_tenancy_case_gateway.rb
mrosel: or an enum - testing on strings is generally not recommended as it is know to cause bugs.

lib/hackney/income/sql_tenancy_case_gateway.rb
mrosel: @madetech-jeffpinkham TIL https://apidock.com/ruby/Enumerable/min_by

madetech-jeffpinkham: Ahhh ok, thanks for explaining that. That makes sense can I suggest renaming it to something like '```UserWithLeastAssignedCases```' or '```UserWithLeastAssignedCases(:priorityBand band)```

madetech-jeffpinkham: Sorry ok, nevermind that's what this method is doing. I'm guessing the min_by is a maths method in a Ruby Library.

madetech-jeffpinkham:
Few minor things but otherwise looks good to me 👍

PR #52, opened 2018-10-11
Pause status

StevenLeighton21:
WHAT:
Add an endpoint that sets the new field on our persisted tenancy model, _paused status_ to the status given for the tenancy ref provided.

WHY:
We will need to use this status to determine if a tenancy should be displayed in a worktray, exempt from automation and potentially even more in future. The IC app will need some way to update this status.

### DIFF
--- a/app/controllers/tenancies_controller.rb
+++ b/app/controllers/tenancies_controller.rb
@@ -5,6 +5,6 @@ class TenanciesController < ApplicationController
       status: params.fetch(:status)
     )
 
-    render :json, :status => 200 if result.nil?
+    render head: :no_content
   end
 end
--- a/app/controllers/tenancies_controller.rb
+++ b/app/controllers/tenancies_controller.rb
@@ -2,9 +2,9 @@ class TenanciesController < ApplicationController
   def update
     result = income_use_case_factory.set_tenancy_paused_status.execute(
       tenancy_ref: params.fetch(:tenancy_ref),
-      status: params.fetch(:status)
+      status: params.fetch(:is_paused)
     )
 
-    render head: :no_content
+    return head(:no_content)
   end
 end
--- a/app/controllers/tenancies_controller.rb
+++ b/app/controllers/tenancies_controller.rb
@@ -1,0 +1,10 @@
+class TenanciesController < ApplicationController
+  def update
+    result = income_use_case_factory.set_tenancy_paused_status.execute(
+      tenancy_ref: params.fetch(:tenancy_ref),
+      status: params.fetch(:status)
+    )
+
+    render :json, :status => 200 if result.nil?
+  end
+end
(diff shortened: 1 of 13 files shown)


app/models/hackney/income/models/tenancy.rb
```diff
@@ -3,6 +3,10 @@ module Income
     module Models
       class Tenancy < ApplicationRecord
         belongs_to :assigned_user, class_name: 'Hackney::Income::Models::User', optional: true
+
+        def paused?
+          self.pause_status
```
mrosel: Best practice kinda thing... use self for setters not getters. 
Actually, this whole method is not needed as AR handles booleans. 

`pause_status?` is already implemented. Can you just change the column to paused?

app/controllers/tenancies_controller.rb
```diff
@@ -0,0 +1,10 @@
+class TenanciesController < ApplicationController
+  def update
+    result = income_use_case_factory.set_tenancy_paused_status.execute(
+      tenancy_ref: params.fetch(:tenancy_ref),
+      status: params.fetch(:status)
+    )
+
+    render :json, :status => 200 if result.nil?
```
mrosel: `204` is what you want here because we aren't returning a response, `200` is expecting a body. 

railsy way would be `  head :no_content `

config/routes.rb
```diff
@@ -2,4 +2,5 @@
   get '/api/v1/my-cases', to: 'my_cases#index'
   get '/api/v1/sync-cases', to: 'my_cases#sync'
   post '/api/v1/users/find-or-create', to: 'users#create'
+  post '/api/v1/tenancies/set-pause-status', to: 'tenancies#update'
```
mrosel: `POST` is a create method `PATCH` and `PUT` are the appropriate restful conventions.

IMO, this needs cleaned up to be nested under api/v1.

app/models/hackney/income/models/tenancy.rb
StevenLeighton21: I had wondered this, and put the helper in without thinking about it. I though `is_paused` might be a good option? Works for the column and the `?` version

app/controllers/tenancies_controller.rb
StevenLeighton21: 👍

config/routes.rb
StevenLeighton21: My bad - completely switched PUT and POST in my head. Use PATCH in this case?

cob16: How about PATCH to '/api/v1/tenancy/:id' with the field name `:paused_date` ? 

This makes adding more editable fields in the future straightforward

spec/controllers/tenancies_controller_spec.rb
```diff
@@ -0,0 +1,87 @@
+require 'rails_helper'
+
+describe TenanciesController do
```
mrosel: you can engage the controller helpers with ,
 ` describe TenanciesController, type: :controller do`

PR #53, opened 2018-10-12
Only return cases that are in arrears.

StevenLeighton21:
Consider cases valid and assigned if they are in arrears, reflect this in total count assigned to user. When returning the worktray, skip cases that are not in arrears.

### DIFF
--- a/lib/hackney/income/dangerous_view_my_cases.rb
+++ b/lib/hackney/income/dangerous_view_my_cases.rb
@@ -17,7 +17,6 @@ module Hackney
         cases = assigned_tenancies.map do |assigned_tenancy|
           tenancy = full_tenancies.find { |t| t.fetch(:ref) == assigned_tenancy.fetch(:tenancy_ref) }
           next if tenancy.nil?
-          next if assigned_tenancy.fetch(:balance) < 0
 
           build_tenancy_list_item(tenancy, assigned_tenancy)
         end.compact
--- a/lib/hackney/income/dangerous_view_my_cases.rb
+++ b/lib/hackney/income/dangerous_view_my_cases.rb
@@ -17,6 +17,7 @@ module Hackney
         cases = assigned_tenancies.map do |assigned_tenancy|
           tenancy = full_tenancies.find { |t| t.fetch(:ref) == assigned_tenancy.fetch(:tenancy_ref) }
           next if tenancy.nil?
+          next if assigned_tenancy.fetch(:balance) < 0
 
           build_tenancy_list_item(tenancy, assigned_tenancy)
         end.compact
--- a/lib/hackney/income/stored_tenancies_gateway.rb
+++ b/lib/hackney/income/stored_tenancies_gateway.rb
@@ -39,7 +39,7 @@ module Hackney
 
       def get_tenancies_for_user(user_id:, page_number: nil, number_per_page: nil)
         query = Hackney::Income::Models::Tenancy
-          .where(assigned_user_id: user_id)
+          .where('tenancies.assigned_user_id = ? AND tenancies.balance > 0', user_id)
           .order(by_band_then_score)
 
         if page_number.present? && number_per_page.present?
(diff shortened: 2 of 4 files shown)


lib/hackney/income/dangerous_view_my_cases.rb
```diff
@@ -17,6 +17,7 @@ def execute(user_id:, page_number:, number_per_page:)
         cases = assigned_tenancies.map do |assigned_tenancy|
           tenancy = full_tenancies.find { |t| t.fetch(:ref) == assigned_tenancy.fetch(:tenancy_ref) }
           next if tenancy.nil?
+          next if assigned_tenancy.fetch(:balance) < 0
```
zuren: Are we consciously making a quick fix, not a proper fix in the API? This will break pagination.

lib/hackney/income/stored_tenancies_gateway.rb
```diff
@@ -50,7 +50,7 @@ def get_tenancies_for_user(user_id:, page_number: nil, number_per_page: nil)
       end
 
       def number_of_pages_for_user(user_id:, number_per_page:)
-        user_cases = Hackney::Income::Models::Tenancy.where(assigned_user_id: user_id)
+        user_cases = Hackney::Income::Models::Tenancy.where("tenancies.assigned_user_id = #{user_id} AND tenancies.balance > 0")
```
zuren: Better practice to use parameterisation here - `where("tenancies.assigned_user_id = ? AND tenancies.balance > 0", user_id)`

Helps protect against SQL injection.

spec/lib/hackney/income/stored_tenancies_gateway_spec.rb
```diff
@@ -268,7 +268,8 @@
     subject { gateway.number_of_pages_for_user(user_id: user_id, number_per_page: number_per_page) }
 
     context 'and the user has ten tenancies' do
-      before { 10.times { create_tenancy(user_id: user_id) } }
+      before { 10.times { create_tenancy(user_id: user_id, balance: 1) } }
+      before { 10.times { create_tenancy(user_id: user_id, balance: -1) } }
```
zuren: Should probably change the context to include "ten tenancies in arrears and ten in credit" and have a more explicit spec around the credit ones not being factored into pagination. It's testing indirectly presently.

Also maybe a bit more readable as the following?

```rb
before do
  10.times { create_tenancy(user_id: user_id, balance: 1) }
  10.times { create_tenancy(user_id: user_id, balance: 1) }
end
```

lib/hackney/income/dangerous_view_my_cases.rb
StevenLeighton21: Why will this break pagination? We're not intending to remove the case from the sync or re-assign it, the intention is that the case is still assigned to the user if it goes back into arrears.

StevenLeighton21: gotcha, thanks!

lib/hackney/income/stored_tenancies_gateway.rb
```diff
@@ -50,7 +50,7 @@ def get_tenancies_for_user(user_id:, page_number: nil, number_per_page: nil)
       end
 
       def number_of_pages_for_user(user_id:, number_per_page:)
-        user_cases = Hackney::Income::Models::Tenancy.where(assigned_user_id: user_id)
+        user_cases = Hackney::Income::Models::Tenancy.where('tenancies.assigned_user_id = ? AND tenancies.balance > 0', user_id)
         (user_cases.count.to_f / number_per_page).ceil
       end
```
zuren: Looks good 👍 

Consider moving `Hackney::Income::Models::Tenancy.where('tenancies.assigned_user_id = ? AND tenancies.balance > 0', user_id)` into a private method as it's used in both `get_tenancies_for_user` and `number_of_pages_for_user`, and it would cause awkward bugs if the query was out of sync between them.

lib/hackney/income/dangerous_view_my_cases.rb
zuren: The StoredTenanciesGateway change looks good, so you should be able to remove the `next if assigned_tenancy.fetch(:balance) < 0` condition here now.

StevenLeighton21:
Merging this - the short lived circle issues have exposed some other inconsistencies in the test that will be resolved outside this PR

PR #54, opened 2018-10-15
Disable scheduling of Sync Cases job

zuren:


StevenLeighton21:
Should x out the relevant test, also :)

StevenLeighton21:
Yeah this is definitely a monitor and fix - there's cards up for limiting the sync

PR #58, opened 2018-10-17
Migrate lcw worktray

StevenLeighton21:
Added a rake task which allows manual migration of synced tenancies in a given patch to a given user ID.

Usage: `rake income:migrate_lcw_cases['patch',user_id]`

### DIFF
--- a/lib/hackney/income/sql_legal_cases_gateway.rb
+++ b/lib/hackney/income/sql_legal_cases_gateway.rb
@@ -1,0 +1,30 @@
+module Hackney
+  module Income
+    class SqlLegalCasesGateway
+      def get_tenancies_for_legal_process_for_patch(patch:)
+        query = database[:tenagree]
+
+        query
+          .left_join(:property, prop_ref: :prop_ref)
+            .where(Sequel[:property][:arr_patch] => patch)
+            .where(Sequel[:tenagree][:tenure] => SECURE_TENURE_TYPE)
+            .where(Sequel[:tenagree][:terminated].cast(:integer) => 0)
+            .where(Sequel[:tenagree][:high_action] => LEGAL_STAGES)
+            .select { Sequel[:tenagree][:tag_ref].as(:tag_ref) }
+            .map { |record| record[:tag_ref].strip }
+      end
+
+      private
+
+      LEGAL_STAGES = ['4RS', '5RP', '6RC', '6RO', '7RE'].freeze
+      SECURE_TENURE_TYPE = 'SEC'.freeze
+
+      def database
+        Hackney::UniversalHousing::Client.connection.tap do |db|
+          db.extension :identifier_mangling
+          db.identifier_input_method = db.identifier_output_method = nil
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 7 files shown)


madetech-jeffpinkham:
A few thoughts as to managing load on the database:

1) What's the approximate number of cases per LCW?
2) Do we have any performance metrics for how long this will take to run?
3) Is there a chance we wouldn't get through assigning all of the cases and some cases would be partially synced.

StevenLeighton21:
There's a hard limit on cases in a patch, I can't really estimate how many there would be that should be migrated to a LCW. As this is a one off, manual migration I believe we can just do it out of business hours with minimal risk, once we've tested it on staging.

lib/hackney/income/sql_legal_cases_gateway.rb
```diff
@@ -0,0 +1,30 @@
+module Hackney
+  module Income
+    class SqlLegalCasesGateway
+      def get_tenancies_for_legal_process_for_patch(patch:)
+        query = database[:tenagree]
+
+        query
+          .left_join(:property, prop_ref: :prop_ref)
+            .where(Sequel[:property][:arr_patch] => patch)
+            .where(Sequel[:tenagree][:tenure] => SECURE_TENURE_TYPE)
```
mrosel: This logic, while it works, feels fragmented because we already have https://github.com/LBHackney-IT/lbh-income-api/blob/8c166df125a05c59d2cdb4d4bcdf8004a484e72a/lib/hackney/income/universal_housing_tenancies_gateway.rb#L20
It looks like we are re-writing sql, where we could have definitions and treat it with our own 'ORM'.

lib/hackney/income/sql_legal_cases_gateway.rb
```diff
@@ -0,0 +1,30 @@
...
+            .where(Sequel[:tenagree][:terminated].cast(:integer) => 0)
```
mrosel: I believe this `.cast` will cause this query to skip indexes, I would suggest bench-marking alternatives.

lib/hackney/income/sql_legal_cases_gateway.rb
StevenLeighton21: Yes, we're rewriting some very similar SQL in order to get a different data set. I felt that this should go into it's own gateway as this is for a manual job, and is _slightly_ different to the other gateway, which is used to generate the sync dataset.

spec/lib/hackney/income/sql_legal_cases_gateway_spec.rb
```diff
@@ -5,41 +5,72 @@
...
+    [
+      Faker::Lorem.characters(4),
+      Faker::Lorem.characters(4),
+      Faker::Lorem.characters(4),
+      Faker::Lorem.characters(4),
+      Faker::Lorem.characters(4)
+    ]
+  end
+
+  let(:patch_code) { Faker::Lorem.characters(3) }
```
StevenLeighton21: We talked about this briefly but I think looking at it now, having an extra patch code seems a little odd. I don't have a better suggestion, but suggest you could define this lone patch code inside the context where it is used, to differentiate it? It might make it clearer that it is used to show the patch code in a different context to the test where you want to generate multiple tenancies from different patches.

PR #59, opened 2018-10-17
Filter paused tenancies (gateway and usecase)

cob16:
This is part of the work for view only paused tenancies.

- gateway takes an optional is_paused arg
  - false filters on not paused
  - true filters on only paused (anticipating future paused cases tab in frontend)
- same for usecase

PS I am not entirely happy with the the early returns in the use case howerver this change would fix a bug we am seeing on senchry production (and new tests).

### DIFF
--- a/lib/hackney/income/dangerous_view_my_cases.rb
+++ b/lib/hackney/income/dangerous_view_my_cases.rb
@@ -8,9 +8,22 @@ module Hackney
         @stored_tenancies_gateway = stored_tenancies_gateway
       end
 
-      def execute(user_id:, page_number:, number_per_page:)
-        number_of_pages_for_user = @stored_tenancies_gateway.number_of_pages_for_user(user_id: user_id, number_per_page: number_per_page)
-        assigned_tenancies = @stored_tenancies_gateway.get_tenancies_for_user(user_id: user_id, page_number: page_number, number_per_page: number_per_page)
+      def execute(user_id:, page_number:, number_per_page:, is_paused: nil)
+        number_of_pages_for_user = @stored_tenancies_gateway.number_of_pages_for_user(
+          user_id: user_id,
+          number_per_page: number_per_page,
+          is_paused: is_paused
+        )
+        return Response.new([], 0) if number_of_pages_for_user.zero?
+
+        assigned_tenancies = @stored_tenancies_gateway.get_tenancies_for_user(
+          user_id: user_id,
+          page_number: page_number,
+          number_per_page: number_per_page,
+          is_paused: is_paused
+        )
+        return Response.new([], number_of_pages_for_user) if assigned_tenancies.length.zero?
+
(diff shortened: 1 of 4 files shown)


lib/hackney/income/dangerous_view_my_cases.rb
```diff
@@ -8,9 +8,22 @@ def initialize(tenancy_api_gateway:, stored_tenancies_gateway:)
...
+        )
+        return Response.new([], 0) if number_of_pages_for_user.zero?
+
+        assigned_tenancies = @stored_tenancies_gateway.get_tenancies_for_user(
+          user_id: user_id,
+          page_number: page_number,
+          number_per_page: number_per_page,
+          is_paused: is_paused
+        )
+        return Response.new([], number_of_pages_for_user) if assigned_tenancies.length.zero?
```
StevenLeighton21: Can `number_of_pages_for_user` ever be anything other than 0 if the user has no assigned cases?

spec/lib/hackney/income/stored_tenancies_gateway_spec.rb
```diff
@@ -189,18 +191,18 @@
...
+                                       ))
           end
         end
 
         context 'and the cases are assigned different bands and scores' do
           let(:multiple_attributes) do
             [
-              { tenancy_ref: Faker::Internet.slug, priority_band: 'red', priority_score: 1, balance: 1},
-              { tenancy_ref: Faker::Internet.slug, priority_band: 'green', priority_score: 50, balance: 1},
+              { tenancy_ref: Faker::Internet.slug, priority_band: 'red', priority_score: 1, balance: 1 },
```
StevenLeighton21: Props for the general cleanup and indenting work 👍 
I think we should get rubocop in here...

spec/lib/hackney/income/stored_tenancies_gateway_spec.rb
```diff
@@ -313,8 +315,88 @@
...
   end
 
-  def create_tenancy(user_id: nil, balance: 1)
-    Hackney::Income::Models::Tenancy.create(assigned_user_id: user_id, balance: balance)
+  context 'there are paused and paused tenancies' do
+    let(:is_paused) { nil }
+    let(:user_id) { Faker::Number.number(2).to_i }
+
+    before do
+      10.times do
```
StevenLeighton21: Suggest that since the _number_ of paused cases is what is being tested, you could triangulate this but having a random number of paused and unpaused cases.
```ruby
let(:num_pased_cases) { Faker::Number.between(1, 10) }

...

num_pased_cases.times do #create paused case end

...


expect(subject.count).to eq(num_pased_cases)
```

StevenLeighton21: There isn't a precedent for this in the existing suite, but it occurs to me that it's something we could do moving forward :)

StevenLeighton21:
Looks great to me 👍 I approve, just allowing some time for others to review

PR #61, opened 2018-10-18
Filter paused tenancies (route)

cob16:


app/controllers/my_cases_controller.rb
```diff
@@ -1,14 +1,28 @@
...
+  before_action :sanitize_params
+
+  REQUIRED_INDEX_PARAMS = %i[user_id page_number number_per_page].freeze
+
   def index
     response = income_use_case_factory.view_my_cases.execute(
-      user_id: params.fetch(:user_id).to_i,
-      page_number: params.fetch(:page_number).to_i,
-      number_per_page: params.fetch(:number_per_page).to_i
+      user_id: view_my_cases_params[:user_id].to_i,
```
elena-vi: Since you have a `sanitize_params` method, wouldn't it be a good idea to move the `.to_i`s to that method too, so your params are ready to be used straight away

app/controllers/my_cases_controller.rb
```diff
@@ -1,14 +1,28 @@
...
+      page_number: view_my_cases_params[:page_number].to_i,
+      number_per_page: view_my_cases_params[:number_per_page].to_i,
+      is_paused: view_my_cases_params[:is_paused]
     )
 
     render json: response
   end
 
+  def view_my_cases_params
```
elena-vi: could we rename this method to follow the [ControllerName_params convention ](https://guides.rubyonrails.org/action_controller_overview.html#strong-parameters)

spec/controllers/my_cases_controller_spec.rb
```diff
@@ -43,6 +47,22 @@
...
         expect(response.body).to eq(expected_result.to_json)
       end
+
+      it 'should respond with only non paused results when requested' do
+        expected_result = {
+          cases: [Faker::GreekPhilosophers.quote],
+          number_per_page: number_per_page
+        }
+
+        allow_any_instance_of(Hackney::Income::DangerousViewMyCases)
```
mrosel: recommend not using `allow_any_instance_of`, see https://www.rubydoc.info/gems/rubocop-rspec/1.7.0/RuboCop/Cop/RSpec/AnyInstance

PR #63, opened 2018-10-23
Case should pause for a determined amount of time

elena-vi:


elena-vi:
[Case should pause for a determined amount of time](https://trello.com/c/PFryfoMy/35-case-should-pause-for-a-determined-amount-of-time)

spec/lib/hackney/income/set_tenancy_paused_status_spec.rb
```diff
@@ -3,22 +3,23 @@
...
 
-        subject.execute(tenancy_ref: tenancy_ref, status: true)
+        subject.execute(tenancy_ref: tenancy_ref, until_date: future_date)
       end
     end
 
     context 'when the operation is unsuccessful' do
       subject { described_class.new(gateway: PauseGatewayDouble.new(true)) }
 
       it 'should not catch the exception raised' do
```
StevenLeighton21: I think another test covering the exception that can be raised if the input fails to parse into a date would be valuable.
We can be reasonably sure our front end will only ever submit something that can be parsed, and that if it's not catching one exception it's _probably_ not catching the other, but it seems like behaviour that would benefit from a test if it ever gets inadvertently changed.

lib/hackney/income/stored_tenancies_gateway.rb
```diff
@@ -56,9 +56,16 @@ def number_of_pages_for_user(user_id:, number_per_page:, is_paused: nil)
...
+        query = Hackney::Income::Models::Tenancy.where('
+          tenancies.assigned_user_id = ? AND
+          tenancies.balance > 0', user_id)
+
+        return query if is_paused.nil?
+
+        if is_paused
+          query = query.where('is_paused_until >= ?', Date.today)
+        else
+          query = query.where('is_paused_until < ? OR is_paused_until is null', Date.today)
```
mrosel: It's not clear the intent here, suggest that we clarify this behavior. 
If I read my intent, I want tenancies that are paused, but if I want tenancies that aren't, this isn't covered -
this query is reflecting tenancies that were paused and no longer are, or never were. I would expect this to return paused tenancies, or all tenancies

lib/hackney/income/sql_pause_tenancy_gateway.rb
```diff
@@ -1,10 +1,15 @@
 module Hackney
   module Income
     class SqlPauseTenancyGateway
-      def set_paused_status(tenancy_ref:, status:)
+      def set_paused_until(tenancy_ref:, until_date:)
```
StevenLeighton21: We now have multiple operations that could raise an error but we raise a custom error in one case, but let the library raise the error in another (update!).I _think_ a more common pattern now would be something like
```ruby
begin
  do dangerous thing1
  do dangerous thing2
rescue FirstBadThing
  raise 'bad thing 1 happened'
rescue SecondBadThing
 raise 'bad thing 2 happened'
rescue AnythingElse
  raise 'something entirely unexpected happened'
end```

File: lib/hackney/income/stored_tenancies_gateway.rb (hunk 2 above)
Comment (mrosel) [reply lib/hackney/income/stored_tenancies_gateway.rb]: you can actually shorten this whole method by
```ruby
def cases_for(user_id, is_paused)
  query = Hackney::Income::Models::Tenancy.where('
    tenancies.assigned_user_id = ? AND
    tenancies.balance > 0', user_id)


  query.where('is_paused_until >= ?', Date.today) if is_paused
end
```

PR #64, opened 2018-10-25
Add linting

zuren:


cob16:
Nice work 👍 

We should add `make lint` and `make check` so it matches the frontend

zuren:
Good point @cob16 - thanks! Added.

mrosel:
Looks good and don't see anything that would change behavior.

PR #67, opened 2018-10-26
Mr guard

mrosel:
`make guard` will bring up the guard cli and automatically run with file changes

1. Runs tests on file change
2. Randomizes order for tests (warning, this may cause some current tests to fail)
3. uses `rubocop -a` to automatically lint and make appropriate changes


mrosel:
It's like running rspec in the console - only difference from docker vs local is the notifications. You won't get notifications from docker.

PR #69, opened 2018-10-29
add brakeman

mrosel:
Add's brakeman to generate a report after full spec runs.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -98,4 +98,5 @@ RSpec.configure do |config|
   # as the one that triggered the failure.
   Kernel.srand config.seed
 =end
+  config.after(:suite) { Brakeman.run app_path: ".", summary: true }
 end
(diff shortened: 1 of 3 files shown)


spec/spec_helper.rb
```diff
@@ -98,4 +98,5 @@
   # as the one that triggered the failure.
   Kernel.srand config.seed
 =end
+  config.after(:suite) { Brakeman.run app_path: ".", summary: true }
```
zuren: Quite like running brakeman separately from specs, i.e.

```
$ brakeman
```

Could add this to `.circleci/config.yml` and `make check`.

That way our TDD cycle isn't slowed down, and the circle workflow is a bit easier to read IMO.

mrosel: This would pull it forward and prevent us from even committing code with known vulnerabilities. Relying on circle, while I think a good suggestion, means that we may not catch an issue until a PR is raised. To me, that's too late in the cycle.

PR #83, opened 2018-11-29
add new messages action diary action codes

cob16:


spec/lib/hackney/income/send_email_spec.rb
```diff
@@ -60,7 +60,7 @@
       .with(
         user_id: user_id,
         tenancy_ref: tenancy_1.tenancy_ref,
-        action_code: 'GAE',
+        action_code: 'GME',
```
mrosel: Suggest, since we're refactoring this, to move this to a variable - the test case should apply to all manual emails, 
Also, this is in an `xit` block and isn't actually testing the use case.

spec/lib/hackney/income/send_sms_spec.rb
```diff
@@ -72,7 +72,7 @@
       .with(
         user_id: user_id,
         tenancy_ref: tenancy.tenancy_ref,
-        action_code: 'GAT',
+        action_code: 'GMS',
```
mrosel: same as ^ https://github.com/LBHackney-IT/lbh-income-api/pull/83/files#diff-e97d828b612c977593844e11cd671d5aR63

PR #84, opened 2018-11-30
Api version in app and action diary

cob16:
- This moves more of the tenancy API path into the app
   - in particular, the '/api/v1' part so that v2 routes can be accessed
   - the app now uses TENANCY_API_HOST and TENANCY_API_KEY (the old env vars can be removed once deployed)

- This fixes the ActionDiaryGateway that relied on using v2

### DIFF
--- a/spec/lib/hackney/tenancy/action_diary_gateway_spec.rb
+++ b/spec/lib/hackney/tenancy/action_diary_gateway_spec.rb
@@ -27,7 +27,7 @@ describe Hackney::Tenancy::Gateway::ActionDiaryGateway do
       )
 
       assert_requested(
-        :post, host + '/tenancies/arrears-action-diary',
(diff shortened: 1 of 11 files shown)


spec/lib/hackney/tenancy/action_diary_gateway_spec.rb
```diff
@@ -9,16 +9,22 @@
...
+      'Accept': 'application/json'
+    }
+  end
 
   subject { described_class.new(host: host, api_key: key) }
 
   context 'when creating an action diary entry' do
     before do
-      stub_request(:post, /#{host}/).with(headers: { API_HEADER_NAME => key }).to_return(status: 200)
+      stub_request(:post, /#{host}/).with(headers: required_headers).to_return(status: 200)
```
elena-vi: YASS DOWN WITH THOSE HASHROCKETS 🔥 🚀

PR #86, opened 2018-12-04
Mr delayed job sync threads

mrosel:
- Removes some bloated files that were complicated and duplicating. 
- Changes behavior to follow delayed_job documentation
- Extends the framework to support tasks, no longer needing complicated docker containers and separate running applications
- deprecates the old sync tasks
- cleans some of the naming


config/application.rb
```diff
@@ -16,6 +16,7 @@
 # Require the gems listed in Gemfile, including any gems
 # you've limited to :test, :development, or :production.
 Bundler.require(*Rails.groups)
+ActiveSupport::Deprecation.behavior = :stderr
```
cob16: Will this give the correct logging output? 
```suggestion
ActiveSupport::Deprecation.behavior = :log
```

config/puma.rb
```diff
@@ -4,7 +4,7 @@
 # the maximum value specified for Puma. Default is set to 5 threads for minimum
 # and maximum; this matches the default thread size of Active Record.
 #
-threads_count = ENV.fetch('RAILS_MAX_THREADS') { 5 }
+threads_count = ENV.fetch('RAILS_MAX_THREADS') { 3 }
```
cob16: whats the motivation for the reduction?

cob16:
>     * Extends the framework to support tasks, no longer needing complicated docker containers and separate running applications

What does this mean? don't we still require the separe job worker containers?

config/application.rb
mrosel: ah yeah, good call. This is already covered in with `config.active_support.deprecation` in the environment files, so I removed it.

mrosel:
You'll still need the workers, but you don't need long running containers for scheduled tasks.

config/puma.rb
mrosel: It's not entirely necessary but works better(slightly) with out resource constraints, since this is deployed on shared vm's with only 2 cores.

PR #88, opened 2018-12-06
Green in arrears job

cob16:
#### Renames
- rename get_tenancies_for_message_1 to 'criteria_for_green_in_arrears'
- rename send_x to manual_send_x in antiaption of createing automated_send_x usecases

#### New
- Create a ContactsGateway that queries the tenancy API /contacts route and returns a new domain object 'Contacts'


lib/hackney/tenancy/gateway/contacts_gateway.rb
```diff
@@ -0,0 +1,49 @@
...
+          json_to_domain_contact_array(response.body)
+        end
+
+        def json_to_domain_contact_array(json_string, responsible_only: true)
+          json = JSON.parse(json_string, symbolize_names: true)
+          json = json.dig(:data, :contacts)
+          json = json.select { |contact| contact[:responsible] } if responsible_only
+
+          contacts = []
+          json.each do |contact|
```
mrosel: suggest cleaning this up. Instead of enumerating and using a nested loop - how about with `.each_with_object`  
```
json.each_with_object([]) do |contact, contacts|

  phone_numbers = json.each_with_object([]) do |contact, phones| 
    phones << contact[:telephone1] if contact[:telephone1]
    phones << contact[:telephone2] if contact[:telephone2]
    phones << contact[:telephone3] if contact[:telephone3]
  end

  contacts << Hackney::Income::Domain::Contact.new.tap do |c|
              c.phone_numbers = phone_numbers
              c.email = contact[:email]
            end  
end
```

spec/lib/hackney/tenancy/gateway/contacts_gateway_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+    end
+
+    it 'should return an array of Hackney::Income::Domain::Contact objects' do
+      expect(subject).to all(be_an(Hackney::Income::Domain::Contact))
+    end
+
+    it 'should return an array of available phone numbers' do
+      expect(subject.first.phone_numbers).to eq(
+        [
+           example_return[:data][:contacts].first[:telephone1],
```
mrosel: This might be the time to change this... `telephone1` and `telephone2`, etc, just feels restrictive and couples our design with a bad legacy design. Could we refactor this to have a contact return an array (hash) of telephones?

cob16: this is indeed what it returns

lib/hackney/tenancy/gateway/contacts_gateway.rb
cob16: nice, didn't know about that method

PR #90, opened 2018-12-07
Mr sanitizer

mrosel:
sets up raven according to the docs and gains parity - `Raven 2.7.4 configured not to capture errors: DSN not set` for other environments.

Sanitises params with potentially identifiable data


mrosel:
yes. By default sentry handles the nil case with a nice log message. Very handy.

PR #96, opened 2018-12-20
On automated SMS/Email write to Action Diary

cob16:


lib/hackney/income/gov_notify_gateway.rb
```diff
@@ -30,14 +30,27 @@ def send_email(recipient:, template_id:, reference:, variables:)
         )
       end
 
-      def get_templates(type:)
-        @client.get_all_templates(type: type).collection.map do |template|
-          { id: template.id, name: template.name, body: template.body }
+      def get_template_by_id(id)
+        get_templates&.find { |template_item| template_item[:id] == id }
```
elena-vi: ruby magics.. always baffales me at first glance but makes sense here

PR #99, opened 2019-01-15
parse manual sms phone numbers and use template name in action diary

cob16:


lib/hackney/income/send_manual_sms.rb
```diff
@@ -7,18 +9,26 @@ def initialize(notification_gateway:, add_action_diary_usecase:)
...
+          Rails.logger.info("Manual SMS sent using template_id: #{template_id}, reference was: #{reference}")
+
+          template_name = @notification_gateway.get_template_by_id(template_id)&.fetch(:name) || template_id
+          @add_action_diary_usecase.execute(
+            user_id: user_id,
+            tenancy_ref: tenancy_ref,
+            action_code: Hackney::Tenancy::ActionCodes::MANUAL_SMS_ACTION_CODE,
+            comment: "#{template_name}' SMS sent to '#{phone.full_e164}'"
+          )
+        else
```
mrosel: This `else` doesn't appear to be tested - let's atleast throw a bad number at it for testing purposes.

lib/hackney/income/send_automated_email.rb
```diff
@@ -13,7 +13,7 @@ def execute(tenancy_ref:, recipient:, template_id:, reference:, variables:)
           reference: reference,
           variables: variables
         )
-        template_name = @notification_gateway.get_template_by_id(template_id).fetch(:name)
+        template_name = @notification_gateway.get_template_by_id(template_id)&.fetch(:name) || template_id
```
mrosel: using `try` and it's alias `&.` are generally considered buggy as they lead to unexpected behaviour - similar to raising errors for flow control where the desired outcome isn't explicit. 
Prefer tell don't ask and SUGGEST, you move the logic for template_id into the `get_template_by_id`. You can tell from the smell that you want to return a `:name` from the `get_template_by_id`

cob16: I have moved it all into a get_template_name(id) function.

PR #104, opened 2019-01-23
Rspec liniting

cob16:
- Added the gem 'rubocop-rspec' for rspec based rules.
- Enabled more rspec rules
- Excluded less files

Waiting for @mrosel to get his changes in before merging.


app/models/hackney/income/models/case_priority.rb
```diff
@@ -11,9 +11,9 @@ class CasePriority < ApplicationRecord
         belongs_to :case, class_name: 'Hackney::Income::Models::Case', optional: true
 
         def tenancy_ref
-          ActiveSupport::Deprecation.warn(
-            'Mr Cases says - tenancy_ref has been moved to Hackney::Income::Models::Case'
-          )
+          # ActiveSupport::Deprecation.warn(
```
mrosel: This is actually annoying (I know, because I did it), so lets just add a `TODO:` so we can search for.

PR #105, opened 2019-01-24
Sms content in action diary

cob16:
~Do not merge until #104 is merged~

- GovNotifyGateway returns the "notification_receipt" domain object that contains the body of the sent message

- This is used to create the action diary message

### DIFF
--- a/lib/hackney/income/gov_notify_gateway.rb
+++ b/lib/hackney/income/gov_notify_gateway.rb
@@ -12,22 +12,24 @@ module Hackney
       end
 
       def send_text_message(phone_number:, template_id:, reference:, variables:)
-        @client.send_sms(
+        responce = @client.send_sms(
           phone_number: pre_release_phone_number(phone_number),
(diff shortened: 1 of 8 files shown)


lib/hackney/income/gov_notify_gateway.rb
```diff
@@ -45,6 +47,12 @@ def get_templates(type: nil)
 
       private
 
+      def create_notification_receipt(responce)
+        Hackney::Income::Domain::NotificationReceipt.new.tap do |notification_receipt|
+          notification_receipt.body = responce.content&.fetch('body', nil)
```
mrosel: It would be beneficial to create a `NotificationReceipt` object and pass that around instead of manipulating it's state with tap.. This would eliminate the workarounds (:try) needed if `content` is `nil`

cob16: do you meen by using a initsalizer? What do we return if there realy is nil content?

PR #112, opened 2019-02-06
remove default running and rely on external

mrosel:
Preserves the functionality when using makefile - but won't leave dangling containers on AWS


elena-vi:
screenshot 2019-02-06 at 16 52 57
🤣 🤣 🤣

PR #114, opened 2019-02-14
Service charge data stub

elena-vi:


lib/hackney/service_charge/gateway/service_charge_gateway.rb
```diff
@@ -0,0 +1,60 @@
...
+              arrears_letter_1_date: sc_case.fetch('arrears_letter_1_date'),
+              international: is_international?(sc_case.fetch('correspondence_postcode'))
+            }
+          end
+        end
+
+        private
+
+        def international?(postcode)
+          !UKPostcode.parse(postcode).valid?
```
elena-vi: I really wanted to use `not` here, for readability but rubocop disagreed 🤖

elena-vi:
[Stub Letter Data🔪](https://trello.com/c/KXLvzAJ4/131-stab-letter-data)

spec/lib/hackney/service_charge/gateway/service_charge_gateway_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+      let(:refs) { [] }
+
+      it 'gives no cases' do
+        expect(subject).to be_empty
+      end
+    end
+
+    context 'when the case has a ref and uk correspondence address' do
+      let(:refs) { [456] }
+      let(:test_url) { 'https://example.com/api/v1/cases?tenancy_refs=%5B456%5D' }
```
mrosel: Suggest you move this up a level because rspec will recognize the change in refs and will be easier to understand what's tested.

spec/lib/hackney/service_charge/gateway/service_charge_gateway_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+      end
+
+      it 'recognises international correspondence address' do
+        expect(subject.first[:international]).to eq(false)
+      end
+    end
+
+    context 'when the case has a ref and international correspondence address' do
+      let(:refs) { [123] }
+      let(:test_url) { 'https://example.com/api/v1/cases?tenancy_refs=%5B123%5D' }
```
mrosel: same as ^

spec/support/request_stub_helper.rb
```diff
@@ -0,0 +1,14 @@
+require 'webmock/rspec'
+
+module RequestStubHelper
```
mrosel: this cleans things up... a lot!

.gitignore
```diff
@@ -18,3 +18,4 @@
 !/tmp/.keep
 
 .byebug_history
+.idea/*
```
alexdesi: since this is related to the editor you are using, it's better to have it in your local git ignore Global:
on `~/.gitignore_global`

lib/hackney/service_charge/gateway/service_charge_gateway.rb
```diff
@@ -0,0 +1,60 @@
...
+      class ServiceChargeGateway
+        def initialize(host:, key:)
+          @host = host
+          @key = key
+        end
+
+        def get_cases_by_refs(refs)
+          return [] if refs.empty?
+
+          uri = URI("#{@host}/api/v1/cases?tenancy_refs=#{refs}")
```
alexdesi: better to use a const for the API endpoint `/api/v1/cases`?

lib/hackney/service_charge/gateway/service_charge_gateway.rb
```diff
@@ -0,0 +1,60 @@
...
+
+          req = Net::HTTP::Get.new(uri)
+          req['X-Api-Key'] = @key
```
alexdesi: I am not sure, maybe we can include this in the previous `Net::HTTP::Get.new...`

lib/hackney/service_charge/gateway/service_charge_gateway.rb
```diff
@@ -0,0 +1,60 @@
...
+
+          res = Net::HTTP.start(uri.hostname, uri.port, use_ssl: true) { |http| http.request(req) }
+
+          raise Hackney::Tenancy::Exceptions::TenancyApiException, res unless res.is_a? Net::HTTPSuccess
+
```
alexdesi: suggested small refactor:
```ruby
  uri = URI("#{@host}/api/v1/cases?tenancy_refs=#{refs}")
...
          raise Hackney::Tenancy::Exceptions::TenancyApiException, res unless res.is_a? Net::HTTPSuccess
```

lib/hackney/service_charge/gateway/service_charge_gateway.rb
alexdesi: yep, we need to change that bud cop!

spec/lib/hackney/service_charge/gateway/service_charge_gateway_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+        )
+      end
+
+      it 'uses the host' do
+        subject
+        expect(WebMock).to have_requested(:get, test_url).once
+      end
+    end
+
+    context 'when passing no case refs' do
```
alexdesi: should we get something like 400 error here? (bad request)

lib/hackney/service_charge/gateway/service_charge_gateway.rb
```diff
@@ -0,0 +1,60 @@
...
+          @host = host
+          @key = key
+        end
+
+        def get_cases_by_refs(refs)
+          return [] if refs.empty?
+
+          uri = URI("#{@host}/api/v1/cases?tenancy_refs=#{refs}")
+
+          req = Net::HTTP::Get.new(uri)
```
cob16: Recommend we keep up our convention of using HTTP party just like the contacts gateway.

lib/hackney/service_charge/gateway/service_charge_gateway.rb
elena-vi: why? it's only defined and used here once anyway

alexdesi: I see, but it's better to use a constants because
- they are descriptive: it is easy to understand what is a string for
- we do not want to sprinkle the methods/classes with strings harcoded string and magic numbers
- it's nice to have those  grouped in one places, so with a simple look you can see which endpoints we are using

I would use a constant `API_CASES_ENDPOINT = '/api/v1/cases'` or something like that.

PR #115, opened 2019-02-18
Save docs on cloud

alexdesi:
This PR:

- Add a model Document that stores all the Document refes of the docs uploaded on S3 
- Use the `aws-sdk-s3` gem to store the docs in the cloud
- Create AWS S3 and Fake adaptors
- Add a rake task to save a document

Trello: https://trello.com/c/iEyFdsoR/139-save-letter-1


app/models/document.rb
```diff
@@ -0,0 +1,26 @@
...
+  validates :filename, :uuid, presence: true
+
+  def self.cloud_save(filename)
+    raise "No such file: #{filename}" unless File.exist?(filename)
+
+    uuid = SecureRandom.uuid
+    format = File.extname(filename)
+    new_filename = "#{uuid}.#{format}"
+
+    new_doc = Document.create(filename: filename,
```
mrosel: `Document` is the same name as your model, this will result in a resource clash. Suggest you change it, or if you are intending to use the model then you can replace `Document.create(filename: filename` with `create(filename: filename`

app/models/document.rb
```diff
@@ -0,0 +1,26 @@
+class Document < ApplicationRecord
+  HACKNEY_BUCKET_DOCS = 'hackney-docs'.freeze
+
+  validates :filename, :uuid, presence: true
+
+  def self.cloud_save(filename)
+    raise "No such file: #{filename}" unless File.exist?(filename)
+
+    uuid = SecureRandom.uuid
+    format = File.extname(filename)
```
mrosel: `format` is a reserved word. Please change.

app/models/document.rb
```diff
@@ -0,0 +1,26 @@
+class Document < ApplicationRecord
+  HACKNEY_BUCKET_DOCS = 'hackney-docs'.freeze
+
+  validates :filename, :uuid, presence: true
+
+  def self.cloud_save(filename)
```
mrosel: recommend against using a class method from within a model. This causes the responsibility of `Document` to be too large. Recommend moving this out into a PORO or its own gateway.

app/models/document.rb
```diff
@@ -0,0 +1,26 @@
+class Document < ApplicationRecord
```
elena-vi: Maybe we should follow the nested hackney style for things? 
and have this be something like `Hackney::Models`

config/environments/development.rb
```diff
@@ -45,4 +45,6 @@
   config.file_watcher = ActiveSupport::EventedFileUpdateChecker
 
   config.gov_notify_api_key = 'DevApiKEY12345'
+
+  config.cloud_storage = HackneyCloud::Storage.new(HackneyCloud::Adapter::AwsS3.new)
```
mrosel: IMO development should not be using cloud storage, especially in the same bucket as intended as production. This is compounded  with  with the hard-coded bucket -https://github.com/LBHackney-IT/lbh-income-api/blob/0caa376e7f090ed6966470a5e7b70baebb7f8e13/app/models/document.rb#L2

db/migrate/20190212114317_create_documents.rb
```diff
@@ -0,0 +1,13 @@
+class CreateDocuments < ActiveRecord::Migration[5.2]
+  def change
+    create_table :documents do |t|
+      t.string :uuid
+      t.string :format
```
mrosel: recommend not using a reserved word for a column name.

db/schema.rb
```diff
@@ -69,6 +69,16 @@
     t.index ["priority", "run_at"], name: "delayed_jobs_priority"
   end
 
+  create_table "documents", force: :cascade do |t|
+    t.string "uuid"
+    t.string "format"
+    t.string "metadata"
```
elena-vi: is a string type appropriate here? 🤷‍♀️
I would have used `text` but I'm not sure that's 100% right

lib/hackney_cloud/adapter/aws_s3.rb
```diff
@@ -0,0 +1,17 @@
+module HackneyCloud
```
elena-vi: Same here on the nesting.. 
I'd suggest `Hackney::Cloud::Adapter`

lib/hackney_cloud/adapter/fake.rb
```diff
@@ -0,0 +1,7 @@
+module HackneyCloud
+  module Adapter
+    class Fake
+      def upload(bucketname, filename, new_filename) end
```
mrosel: not sure how this worked without a `;` before `end`?

lib/hackney_cloud/storage.rb
```diff
@@ -0,0 +1,11 @@
+module HackneyCloud
```
elena-vi: Same here on the nesting again.. 
I'd like to make it clear that I'm not super found of it, but let's just make everything consistent

lib/hackney_cloud/adapter/aws_s3.rb
```diff
@@ -0,0 +1,17 @@
...
+  module Adapter
+    class AwsS3
+      def initialize(stub: false)
+        client = Aws::S3::Client.new(region: 'eu-west-2', stub_responses: stub)
```
mrosel: actually, this is a configurable and is already handled for you in the gem - https://github.com/aws/aws-sdk-ruby#configuration

app/models/document.rb
alexdesi: good point, I'll change it

app/models/document.rb
alexdesi: ops 😄 , Iet's use `extension`

app/models/document.rb
alexdesi: agree

app/models/document.rb
alexdesi: I put it inside the `hackney` namespace, not sure if I should create the sub namespace Models, they are already in the models folder.
For example it sounds strange that the income models are stored in the folder:
`/models/hackney/income/models/`
we are repeting models 2 times

alexdesi: any thought on it @mrosel @cob16 ?

config/environments/development.rb
alexdesi: right, my idea is to use the `stub` option available with AWS s3 sdk:
`client = Aws::S3::Client.new(region: 'eu-west-2', stub_responses: stub)`

alexdesi: here is how: https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/ClientStubs.html#stub_responses-instance_method

lib/hackney_cloud/adapter/fake.rb
alexdesi: Yes, it works. If you inline more than one statement that you need the `;`
Anyway I updated this, and now it is multiline

db/schema.rb
alexdesi: we don't know yet the king of metatag we are going to store, the string works well we want to store json

lib/hackney_cloud/adapter/aws_s3.rb
alexdesi: I like it 👍

app/models/hackney/cloud_document.rb
```diff
@@ -0,0 +1,4 @@
+module Hackney
+  class CloudDocument < ApplicationRecord
```
mrosel: looking at previous classes, shouldn't this be `Hackney::Cloud::Document` ?

config/environments/development.rb
```diff
@@ -45,4 +45,6 @@
   config.file_watcher = ActiveSupport::EventedFileUpdateChecker
 
   config.gov_notify_api_key = 'DevApiKEY12345'
+
+  config.cloud_adaptor = HackneyCloud::Adapter::AwsS3.new(stub: true)
```
mrosel: It appears that you meant `Hackney::Cloud` instead of `HackneyCloud`

PR #116, opened 2019-02-19
Letter generator

elena-vi:
[pdf.pdf](https://github.com/LBHackney-IT/lbh-income-api/files/2894189/pdf.pdf)


elena-vi:
[Consume SC stub & generate Letter 1](https://trello.com/c/zMgVHhRE/138-consume-sc-stub-generate-letter-1)

elena-vi:
[Consume SC stub & generate Letter 1](https://trello.com/c/zMgVHhRE/138-consume-sc-stub-generate-letter-1)

lib/hackney/pdf/pdf_generator.rb
```diff
@@ -0,0 +1,28 @@
...
+        lessee_full_name = letter_params.fetch(:lessee_full_name)
+        correspondence_address_one = letter_params.fetch(:correspondence_address_one)
+        correspondence_address_two = letter_params.fetch(:correspondence_address_two)
+        correspondence_address_three = letter_params.fetch(:correspondence_address_three)
+        correspondence_postcode = letter_params.fetch(:correspondence_postcode)
+        lessee_short_name = letter_params.fetch(:lessee_short_name)
+        property_address = letter_params.fetch(:property_address)
+        arrears_letter_1_date = letter_params.fetch(:arrears_letter_1_date)
+        total_collectable_arrears_balance = letter_params.fetch(:total_collectable_arrears_balance)
+
```
alexdesi: we should dry this a bit,
what about just
`params = letter_params` 
and access the params from the ERB through `params`

lib/hackney/pdf/templates/letter_1_template.erb
```diff
@@ -0,0 +1,92 @@
...
+    <path d="M81 425 c-57 -52 -81 -110 -81 -195 0 -50 6 -83 21 -115 20 -43 89 -115 109 -115 6 0 10 35 10 85 l0 85 115 0 115 0 0 -85 c0 -54 4 -85 11 -85 23 0 77 53 105 102 25 46 29 64 29 128 0 88 -19 135 -78 192 -60 58 -67 53 -67 -47 l0 -85 -115 0 -115 0 0 85 c0 50 -4 85 -10 85 -5 0 -28 -16 -49 -35z"/>
+    <path d="M590 225 l0 -215 70 0 70 0 0 75 0 75 55 0 55 0 0 -75 0 -75 70 0 70 0 0 150 0 149 24 -23 c20 -21 60 -34 151 -48 25 -4 31 -33 9 -42 -9 -4 -27 0 -40 9 -13 8 -45 15 -74 15 -46 0 -50 -2 -50 -23 0 -13 12 -34 26 -48 22 -21 37 -24 113 -27 99 -4 147 12 156 51 2 12 6 69 8 127 2 58 7 113 11 123 7 15 1 17 -57 17 -40 0 -67 -4 -69 -11 -3 -9 -9 -9 -25 0 -28 14 -105 14 -137 0 -13 -6 -29 -21 -34 -33 -10 -19 -11 -18 -11 12 l-1 32 -70 0 -70 0 0 -85 0 -85 -55 0 -55 0 0 85 0 85 -70 0 -70 0 0 -215z m588 106 c3 -21 1 -23 -25 -17 -15 4 -32 11 -37 16 -15 14 5 31 32 28 20 -2 28 -10 30 -27z"/>
+  </g>
+</svg>
```
alexdesi: it would be neater if we keep the SGV logo in a different file and import it inside the letters.

lib/hackney/pdf/pdf_gateway.rb
```diff
@@ -0,0 +1,28 @@
+require 'pdfkit'
+
+module Hackney
+  module PDF
+    class PDFGateway
+      def initialize; end
```
mrosel: This is the default and doesn't actually do anything and you can remove it.

lib/hackney/pdf/templates/letter_1_template.erb
```diff
@@ -0,0 +1,92 @@
...
+  </p>
+  <p>
+    We look forward to receiving your payment within 14 days from the date of this letter.
+  </p>
+
+  <p>
+    Yours sincerely,<br>
+    Service Charge Team
+  </p>
+</div>
```
alexdesi: the template is very clean 👍 
about the variables, we can just pass just one hash (`params[:correspondence_address_one]` for example)

lib/hackney/pdf/templates/pdf_styles.css
```diff
@@ -0,0 +1,53 @@
...
+
+.top_space {
+  height: 69.12pt;
+}
+
+.hackney-logo {
+  margin-top: 25.92pt;
+}
+
+
```
alexdesi: 🔍  2 lines :)

lib/hackney/pdf/templates/pdf_styles.css
```diff
@@ -0,0 +1,53 @@
...
+.hackney-lines{
+  width: 319.68pt;
+  margin-left: 286pt;
+  position: absolute;
+  top: 592.12pt;
+  z-index: -1;
+}
+
+
```
alexdesi: 2 empty lines

spec/lib/hackney/pdf/pdf_gateway_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+describe Hackney::PDF::PDFGateway do
+  let(:gateway) { described_class.new }
+  let(:html) { "<h1>#{Faker::RickAndMorty.quote}</h1>" }
+
+  context 'when generating a pdf' do
+    subject do
+      gateway.generate_pdf(html)
+    end
+
+    it 'pdf should have the right options' do
```
alexdesi: let's not use should in our tests, it is better just to describe it using present tense: `pdf has right options`
(see http://www.betterspecs.org/ for more)

spec/lib/hackney/pdf/pdf_generator_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+      template_path: test_template_path,
+      pdf_gateway: pdf_gateway
+    )
+  end
+
+  let(:pdf_gateway) { instance_double(Hackney::PDF::PDFGateway) }
+  let(:test_template_path) { 'spec/lib/hackney/pdf/test_template.erb' }
+  let(:test_letter_params) do
+    {
+
```
alexdesi: empty line

lib/hackney/pdf/pdf_generator.rb
elena-vi: I'd like to keep this here, because the is a [card](https://trello.com/c/jjZPk7OR/154-step-2-ui-preview-letter) in our to do list where we will need to highlight fields which are missing so having them there allows us to have more control of what error messages we should introduce later on, also this allows us to keep the templates not tied to any data structure.

alexdesi: Thats a very good point.

> allows us to have more control of what error messages we should introduce later on

Let's be more explicit then:
let's add a constant:
```
      REQUIRED_LETTER_PARAMS = [:payment_ref, :lessee_full_name, :correspondence_address_one, :correspondence_address_two,
          :correspondence_address_three, :correspondence_postcode, :lessee_short_name,
          :property_address, :arrears_letter_1_date, :total_collectable_arrears_balance]
```

and validate the params:
```
      def validate_params(letter_params)
        REQUIRED_LETTER_PARAMS.each do |param|
          raise "#{param} is not present!" unless letter_params.include?(param)
        end
      end
```

this simply the `execute` method:
something like this:

```
      def execute(letter_params:)
        validate_params(letter_params)

        params = letter_params

        template = File.open(@template_path).read
        html = ERB.new(template).result(binding)

        @pdf_gateway.generate_pdf(html)
      end
```

alexdesi: > also this allows us to keep the templates not tied to any data structure

If we do not pass a Hash, the templates will be tied to the variable, and that is bad code smell.
Using a hash in the template is much better that using lot's of variable with different names.

elena-vi: this is great but the [4th AC](https://trello.com/c/jjZPk7OR/154-step-2-ui-preview-letter) mentions that we still need to render the html with the missing fields.. I'd like to keep this in for now and address it later when we have a better feel for the user flow

lib/hackney/pdf/templates/letter_1_template.erb
elena-vi: I'd like to leave that in for now and then move it out later because we'll need this in more than one file

lib/hackney/pdf/pdf_generator.rb
alexdesi: Sure, address this later then 👍 !
(superhappy to pair with you on it if you want)

PR #121, opened 2019-03-04
Mr redis cache jobs

mrosel:
Extends the application to use sidekiq with redis
Removes a bunch of redundant tests and code.


config/application.rb
```diff
@@ -36,7 +36,8 @@ class Application < Rails::Application
 
     config.eager_load_paths << Rails.root.join('lib')
 
-    config.active_job.queue_adapter = :delayed_job
+    # config.active_job.queue_adapter = :delayed_job
```
elena-vi: probably don't need this here...

config/initializers/sidekiq.rb
```diff
@@ -0,0 +1,14 @@
+redis_url = ENV.fetch('REDIS_URL')
+
+Sidekiq.configure_server do |config|
+  # config.redis = { url: "redis://redis:6379" }
```
elena-vi: this too...

app/jobs/application_job.rb
```diff
@@ -6,30 +6,4 @@ def income_use_case_factory
...
-      set(wait_until: next_run_time).perform_later
-    end
-
-    def next_run_time
-      raise NotImplementedError
-    end
-
-    private
-
-    def already_queued_for_next_run?
```
elena-vi: this looks hella confusing and smells like we're reinventing how jobs are supposed to be enqueued. I'm happy to see this go 🔥

PR #125, opened 2019-03-07
web for Sidekiq and scheduler

mrosel:
dashboard/sidekiq under basic auth

### DIFF
--- a/config/routes/sidekiq.rb
+++ b/config/routes/sidekiq.rb
@@ -7,12 +7,11 @@ Sidekiq::Web.use Rack::Auth::Basic do |username, password|
   # - See https://thisdata.com/blog/timing-attacks-against-string-comparison/
   # - Use & (do not use &&) so that it doesn't short circuit.
   # - Use digests to stop length information leaking (see also ActiveSupport::SecurityUtils.variable_size_secure_compare)
-  ActiveSupport::SecurityUtils.secure_compare(::Digest::SHA256.hexdigest(username), ::Digest::SHA256.hexdigest(ENV["SIDEKIQ_USERNAME"])) &
-    ActiveSupport::SecurityUtils.secure_compare(::Digest::SHA256.hexdigest(password), ::Digest::SHA256.hexdigest(ENV["SIDEKIQ_PASSWORD"]))
-end # unless Rails.env.development?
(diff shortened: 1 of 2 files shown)


config/routes/sidekiq.rb
```diff
@@ -0,0 +1,18 @@
...
+  # - See https://codahale.com/a-lesson-in-timing-attacks/
+  # - See https://thisdata.com/blog/timing-attacks-against-string-comparison/
+  # - Use & (do not use &&) so that it doesn't short circuit.
+  # - Use digests to stop length information leaking (see also ActiveSupport::SecurityUtils.variable_size_secure_compare)
+  ActiveSupport::SecurityUtils.secure_compare(::Digest::SHA256.hexdigest(username), ::Digest::SHA256.hexdigest(ENV['SIDEKIQ_USERNAME'])) &
+    ActiveSupport::SecurityUtils.secure_compare(::Digest::SHA256.hexdigest(password), ::Digest::SHA256.hexdigest(ENV['SIDEKIQ_PASSWORD']))
+end
+
+Rails.application.routes.draw do
+  scope '/dashboard' do
```
alexdesi: should we use something like `/admin` here,
to be more explicit that you are in admin area here?

config/routes/sidekiq.rb
```diff
@@ -0,0 +1,18 @@
+require 'sidekiq/web'
+require 'sidekiq-scheduler/web'
+
+Sidekiq::Web.use Rack::Auth::Basic do |username, password|
+  # Protect against timing attacks:
+  # - See https://codahale.com/a-lesson-in-timing-attacks/
+  # - See https://thisdata.com/blog/timing-attacks-against-string-comparison/
+  # - Use & (do not use &&) so that it doesn't short circuit.
+  # - Use digests to stop length information leaking (see also ActiveSupport::SecurityUtils.variable_size_secure_compare)
```
alexdesi: do we need these comments?

mrosel: don't need them. Since it's dealing with auth, comments can be helpful to future self from making changes. Or classic `'string' == 'string'`. Can remove them if that's a team preference but this hasn't been determined.

PR #128, opened 2019-03-12
Add support for KMS and client-side encryption

alexdesi:
This PR add the Client-Side encryption and support for KMS
(to know how it works read: https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingClientSideEncryption.html)

Trello: https://trello.com/c/MfQ2yWRX/152-kms-envelope-encryption-implementation

### DIFF
--- a/config/cloud_storage.yml
+++ b/config/cloud_storage.yml
@@ -1,11 +1,15 @@
 test:
   bucket_docs: hackney-docs-test
+  customer_managed_key: '-'
 
 development:
   bucket_docs: hackney-docs-development
+  customer_managed_key: <%= ENV['CUSTOMER_MANAGED_KEY'] %>
 
 production:
   bucket_docs: hackney-docs-production
+  customer_managed_key: <%= ENV['CUSTOMER_MANAGED_KEY'] %>
 
 staging:
   bucket_docs: hackney-docs-staging
+  customer_managed_key: <%= ENV['CUSTOMER_MANAGED_KEY'] %>
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -46,5 +46,7 @@ Rails.application.configure do
 
   config.gov_notify_api_key = 'DevApiKEY12345'
 
-  config.cloud_adapter = Hackney::Cloud::Adapter::AwsS3.new(stub: true)
+  # Configure the Cloud storage service
+  encryption_client = Hackney::Cloud::EncryptionClient.new(config_for('cloud_storage')['customer_managed_key']).create
+  config.cloud_adapter = Hackney::Cloud::Adapter::AwsS3.new(encryption_client)
 end
--- a/lib/hackney/cloud/adapter/aws_s3.rb
+++ b/lib/hackney/cloud/adapter/aws_s3.rb
@@ -2,18 +2,33 @@ module Hackney
   module Cloud
     module Adapter
       class AwsS3
-        def initialize(stub: false)
-          client = Aws::S3::Client.new(stub_responses: stub)
-
-          @s3 = Aws::S3::Resource.new(client: client)
+        def initialize(encryption_client)
+          @client = encryption_client
         end
 
-        def upload(bucketname, filename, new_filename)
-          s3_file_obj = @s3.bucket(bucketname).object(new_filename)
+        def upload(bucket_name:, filename:, new_filename:)
+          content = File.read(filename)
 
-          success = s3_file_obj.upload_file(filename)
+          # Add encrypted item to bucket
+          resp = client.put_object(
+            body: content,
+            bucket: bucket_name,
+            key: new_filename
+          )
 
-          success ? "#{s3_file_obj.public_url}/#{new_filename}" : nil
+          resp.successful? || raise('Cloud Storage Error!')
+        end
+
+        def download(bucket_name, filename)
+          client.get_object(bucket: bucket_name, key: filename)
+        end
+
+        private
+
+        attr_reader :client
+
+        def customer_managed_key
+          Rails.application.config_for('cloud_storage')['customer_managed_key']
         end
       end
     end
--- a/spec/lib/hackney/cloud/adapter/aws_s3_spec.rb
+++ b/spec/lib/hackney/cloud/adapter/aws_s3_spec.rb
@@ -1,14 +1,51 @@
 require 'rails_helper'
 
 describe Hackney::Cloud::Adapter::AwsS3 do
-  subject(:s3) { described_class.new(stub: true) }
+  subject(:s3) { described_class.new(encryption_client_double) }
+
+  let(:encryption_client_double) { double }
+
+  let(:new_filename) { 'new_filename.txt' }
 
   let(:filename) { './spec/lib/hackney/cloud/adapter/upload_test.txt' }
-  let(:new_filename) { SecureRandom.uuid }
+  let(:content) { File.read(filename) }
 
-  it 'uploads a file on S3' do
-    s3_url = s3.upload('my-bucket', filename, new_filename)
+  context 'when there are NOT upload errors' do
+    let(:upload_response) { double(successful?: true) }
 
-    expect(s3_url).to match `https:\/\/my-bucket.*#{new_filename}`
+    it 'successfully uploads the S3' do
+      allow(encryption_client_double).to receive(:put_object)
+        .with(body: content, bucket: 'my-bucket', key: new_filename)
+        .and_return(upload_response)
+
+      expect(
+        s3.upload(bucket_name: 'my-bucket',
+                  filename: filename,
+                  new_filename: new_filename)
+      ).to be true
+    end
+  end
+
+  context 'when there are upload errors' do
+    let(:upload_response) { double(successful?: false) }
+
+    it 'raises an exception' do
+      allow(encryption_client_double).to receive(:put_object)
+        .with(body: content, bucket: 'my-bucket', key: new_filename)
+        .and_return(upload_response)
+
+      expect {
+        s3.upload(bucket_name: 'my-bucket',
+                  filename: filename,
+                  new_filename: new_filename)
+      }.to raise_exception('Cloud Storage Error!')
+    end
+  end
+
+  it 'downloads a file from S3' do
+    expect(encryption_client_double).to receive(:get_object)
+      .with(bucket: 'my-bucket', key: filename)
+
+    s3.download('my-bucket', filename)
   end
 end
--- a/app/jobs/hackney/cloud/jobs/save_to_cloud_job.rb
+++ b/app/jobs/hackney/cloud/jobs/save_to_cloud_job.rb
@@ -7,7 +7,9 @@ module Hackney
         queue_as :cloud_storage
 
         def perform(bucket_name:, filename:, new_filename:, model_document:, uuid:)
-          url = cloud_provider.upload(bucket_name, filename, new_filename)
+          url = cloud_provider.upload(bucket_name: bucket_name,
+                                      filename: filename,
+                                      new_filename: new_filename)
 
           document(model_document, uuid).update(url: url, status: UPLOADED_CLOUD_STATUS)
         end
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -80,6 +80,7 @@ Rails.application.configure do
 
   config.x.gov_notify.api_key = ENV.fetch('GOV_NOTIFY_API_KEY')
 
-  # This specify the Cloud storage service that we are using
-  config.cloud_adapter = Hackney::Cloud::Adapter::AwsS3.new
+  # Configure the Cloud storage service
+  encryption_client = Hackney::Cloud::EncryptionClient.new(config_for('cloud_storage')['customer_managed_key']).create
+  config.cloud_adapter = Hackney::Cloud::Adapter::AwsS3.new(encryption_client)
 end
--- a/config/environments/staging.rb
+++ b/config/environments/staging.rb
@@ -80,6 +80,7 @@ Rails.application.configure do
 
   config.x.gov_notify.api_key = ENV.fetch('GOV_NOTIFY_API_KEY')
 
-  # This specify the Cloud storage service that we are using
-  config.cloud_adapter = Hackney::Cloud::Adapter::AwsS3.new
+  # Configure the Cloud storage service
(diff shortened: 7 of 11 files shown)


config/cloud_storage.yml
```diff
@@ -1,11 +1,15 @@
...
   bucket_docs: hackney-docs-development
+  customer_managed_key: 40255180-966f-45aa-bd0a-762e0d4b244d
 
 production:
   bucket_docs: hackney-docs-production
+  customer_managed_key: 11f3b9c0-f485-4e90-9c04-4fecd416496b
 
 staging:
   bucket_docs: hackney-docs-staging
+  customer_managed_key: 40255180-966f-45aa-bd0a-762e0d4b244d
```
elena-vi: We probably not be storing these here.... and we'll need to purge our git history of these

lib/hackney/cloud/adapter/aws_s3.rb
```diff
@@ -2,18 +2,40 @@ module Hackney
...
-          s3_file_obj = @s3.bucket(bucketname).object(new_filename)
+        def upload(bucket_name, filename, new_filename)
+          contents = File.read(filename)
 
-          success = s3_file_obj.upload_file(filename)
+          # Add encrypted item to bucket
+          resp = client.put_object(
+            body: contents,
+            bucket: bucket_name,
+            key: new_filename
```
elena-vi: what is this `new_filename` going to be?

lib/hackney/cloud/adapter/aws_s3.rb
```diff
@@ -2,18 +2,40 @@ module Hackney
...
+          # Create encryption client
+          @client = encryption_client || Aws::S3::Encryption::Client.new(
+            kms_key_id: customer_managed_key,
+            kms_client: kms
+          )
         end
 
-        def upload(bucketname, filename, new_filename)
-          s3_file_obj = @s3.bucket(bucketname).object(new_filename)
+        def upload(bucket_name, filename, new_filename)
```
elena-vi: maybe these should be named params, just to make it easier to understand when they are called

lib/hackney/cloud/adapter/aws_s3.rb
```diff
@@ -2,18 +2,40 @@ module Hackney
   module Cloud
     module Adapter
       class AwsS3
-        def initialize(stub: false)
-          client = Aws::S3::Client.new(stub_responses: stub)
+        def initialize(encryption_client: nil, stub: false)
+          # Create KMS client
+          kms = Aws::KMS::Client.new(stub_responses: stub)
```
elena-vi: I'm a little confused on how this `stub_responses` works and where it's defined

lib/hackney/cloud/adapter/aws_s3.rb
```diff
@@ -2,18 +2,40 @@ module Hackney
...
+          @client = encryption_client || Aws::S3::Encryption::Client.new(
+            kms_key_id: customer_managed_key,
+            kms_client: kms
+          )
         end
 
-        def upload(bucketname, filename, new_filename)
-          s3_file_obj = @s3.bucket(bucketname).object(new_filename)
+        def upload(bucket_name, filename, new_filename)
+          contents = File.read(filename)
```
elena-vi: minor thing, but I'm guessing this will just be a singular `content` of the file not multiple contents
```suggestion
          content = File.read(filename)
```

lib/hackney/cloud/adapter/aws_s3.rb
```diff
@@ -2,18 +2,40 @@ module Hackney
...
 
-          success = s3_file_obj.upload_file(filename)
+          # Add encrypted item to bucket
+          resp = client.put_object(
+            body: contents,
```
elena-vi: same as above ☝️ 
```suggestion
            body: content,
```

spec/lib/hackney/cloud/adapter/aws_s3_spec.rb
```diff
@@ -1,14 +1,27 @@
...
-  let(:new_filename) { SecureRandom.uuid }
+  let(:new_filename) { 'new_filename.txt' }
+  let(:encryption_client) { double }
+  let(:contents) { File.read(filename) }
 
   it 'uploads a file on S3' do
-    s3_url = s3.upload('my-bucket', filename, new_filename)
+    expect(encryption_client).to receive(:put_object)
+      .with(body: contents, bucket: 'my-bucket', key: new_filename)
+      .and_return(true)
```
elena-vi: Is the real response from aws going to be just `true`?
I'd probably have this method return something that's a little more _real_ looking

config/cloud_storage.yml
cob16: it might just easier just to make new keys and leave the history as this pull request is already public

lib/hackney/cloud/adapter/aws_s3.rb
mrosel: I was under the impression that we were going to use https://github.com/ankane/kms_encrypted. Why the diversion? IMO `kms_encrypted` covers the use case that we wanted.

lib/hackney/cloud/adapter/aws_s3.rb
```diff
@@ -2,18 +2,40 @@ module Hackney
   module Cloud
     module Adapter
       class AwsS3
-        def initialize(stub: false)
-          client = Aws::S3::Client.new(stub_responses: stub)
+        def initialize(encryption_client: nil, stub: false)
```
mrosel: the need to pass encryption_client explicitly feels unnecessary as we can only use this library with `Aws::S3::Encryption::Client`

lib/hackney/cloud/adapter/aws_s3.rb
```diff
@@ -2,18 +2,40 @@ module Hackney
...
         end
 
-        def upload(bucketname, filename, new_filename)
-          s3_file_obj = @s3.bucket(bucketname).object(new_filename)
+        def upload(bucket_name, filename, new_filename)
+          contents = File.read(filename)
 
-          success = s3_file_obj.upload_file(filename)
+          # Add encrypted item to bucket
+          resp = client.put_object(
```
mrosel: Can you clarify? This is confusing because it looks like we're orphaning the existing document and replacing it with a different document.

lib/hackney/cloud/adapter/aws_s3.rb
```diff
@@ -2,18 +2,40 @@ module Hackney
...
+            body: contents,
+            bucket: bucket_name,
+            key: new_filename
+          )
 
-          success ? "#{s3_file_obj.public_url}/#{new_filename}" : nil
+          resp
```
mrosel: Should remove `resp` as it's an antipattern to set a variable and return it.

config/cloud_storage.yml
alexdesi: Ops, yes sure, I've just disabled the key (I'll remove those),
we'll read the `cmk` from Env

lib/hackney/cloud/adapter/aws_s3.rb
alexdesi: I am also using this for testing purpose,
since the use of `allow_any_instance_of` in testing is not recommended.

lib/hackney/cloud/adapter/aws_s3.rb
alexdesi: It is the way to stub works for testing: 
https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/ClientStubs.html#stub_responses-instance_method

alexdesi: I am not sure, I didn't notice that gem actually.
We are using the official Ruby AWS Sdk, with client side encr + mks support.
Not sure about the advantages of that Gem (also it is basically mantained by just one guy).

lib/hackney/cloud/adapter/aws_s3.rb
alexdesi: - we get a new instance of  Encryption::Client
- and then we call put object on it
Here is how it works: https://docs.aws.amazon.com/sdk-for-ruby/v3/developer-guide/s3-example-client-side-encryption-with-kms-master-key.html

lib/hackney/cloud/adapter/aws_s3.rb
mrosel: I don't see how that's related. We should be using a test double anyway.

lib/hackney/cloud/adapter/aws_s3.rb
mrosel: a test double would also fix this antipattern and keep test code isolated from production.

lib/hackney/cloud/adapter/aws_s3.rb
mrosel: looking at that documentation... here it looks like we are orphaning the original document.  `new_filename` should not be used.

lib/hackney/cloud/adapter/aws_s3.rb
alexdesi: It is the name of the of the file (object) that will be used to store it in the S3 Bucket.
The point is that we do not want to use the original name,
in case of the letter, it will be something like `9762387629876876238476283564.pdf`.

lib/hackney/cloud/adapter/aws_s3.rb
alexdesi: I agree, there was more code before, then I refactored and the resp was a left over :)

lib/hackney/cloud/adapter/aws_s3.rb
mrosel: appears it should be `filename` 
referencing: https://github.com/awsdocs/aws-doc-sdk-examples/blob/master/ruby/example_code/s3/s3_add_cskms_encrypt_item.rb#L52

lib/hackney/cloud/adapter/aws_s3.rb
mrosel: It was referenced in internal documentation, so it was a surprise to not see it here. We should discuss offline whether to use the SDK directly.

lib/hackney/cloud/adapter/aws_s3.rb
alexdesi: mhh, maybe I am missing something?
here is the example in the doc:
```ruby
kms = Aws::KMS::Client.new

# Create encryption client
client = Aws::S3::Encryption::Client.new(
  kms_key_id: key,
  kms_client: kms
)

# Add encrypted item to bucket
client.put_object(
  body: contents,
  bucket: bucket,
  key: item
)

puts 'Added client-side KMS encrypted item ' + item + ' to bucket ' + bucket
```
why we should not use the `new_filename` (= the key) ?

lib/hackney/cloud/adapter/aws_s3.rb
alexdesi: as discussed `new_filename` is the name of the obj will have inside the bucket  (9762387629876876238476283564.pdf for instance)
filename is the local file name

config/environments/development.rb
```diff
@@ -46,5 +46,6 @@
 
   config.gov_notify_api_key = 'DevApiKEY12345'
 
-  config.cloud_adapter = Hackney::Cloud::Adapter::AwsS3.new(stub: true)
+  encryption_client = Hackney::Cloud::EncryptionClient.new(ENV.fetch('CUSTOMER_MANAGED_KEY')).create
```
alexdesi: I think we could find a better way for this (explicit encryption_client creation in the config), but at the moment is good enough. We could add a new card to refactor this.

mrosel: I think this is meant to be pulled from `config/cloud_storage.yml`

lib/hackney/cloud/adapter/aws_s3.rb
cob16: Yeah, I was advocating using that gem from the start. There must have been some miscommunication on not to looking at that gem before implementing our own.

spec/lib/hackney/cloud/adapter/aws_s3_spec.rb
```diff
@@ -1,14 +1,51 @@
...
+  end
 
-  it 'uploads a file on S3' do
-    s3_url = s3.upload('my-bucket', filename, new_filename)
+  it 'downloads a file from S3' do
+    expect(encryption_client_double).to receive(:get_object)
+      .with(bucket: 'my-bucket', key: filename)
 
-    expect(s3_url).to match `https:\/\/my-bucket.*#{new_filename}`
+    s3.download('my-bucket', filename)
```
alexdesi: add the expected response as expectation

config/cloud_storage.yml
```diff
@@ -1,11 +1,15 @@
 test:
   bucket_docs: hackney-docs-test
+  customer_managed_key: '-'
 
 development:
   bucket_docs: hackney-docs-development
+  customer_managed_key: <%= ENV['CUSTOMER_MANAGED_KEY_DEVELOPMENT'] %>
```
mrosel: `ENV['CUSTOMER_MANAGED_KEY']`

config/cloud_storage.yml
```diff
@@ -1,11 +1,15 @@
...
 
 production:
   bucket_docs: hackney-docs-production
+  customer_managed_key: <%= ENV['CUSTOMER_MANAGED_KEY_PRODUCTION'] %>
```
mrosel: ENV['CUSTOMER_MANAGED_KEY']

config/cloud_storage.yml
```diff
@@ -1,11 +1,15 @@
...
 
 staging:
   bucket_docs: hackney-docs-staging
+  customer_managed_key: <%= ENV['CUSTOMER_MANAGED_KEY_STAGING'] %>
```
mrosel: ENV['CUSTOMER_MANAGED_KEY']

config/cloud_storage.yml
mrosel:
```suggestion
  customer_managed_key: <%= ENV['CUSTOMER_MANAGED_KEY'] %>
```

config/cloud_storage.yml
mrosel:
```suggestion
  customer_managed_key: <%= ENV['CUSTOMER_MANAGED_KEY'] %>
```

config/environments/development.rb
alexdesi: yes, using `config_for('cloud_storage')['customer_managed_key']`

PR #129, opened 2019-03-14
Refactoring to income to rent

elena-vi:


elena-vi:
[Tech debt income-api: rename hackney:income to rent.](https://trello.com/c/s3SqPGrx/167-tech-debt-income-api-rename-hackneyincome-to-rent)

app/controllers/tenancies_controller.rb
```diff
@@ -1,6 +1,6 @@
 class TenanciesController < ApplicationController
   def update
-    income_use_case_factory.set_tenancy_paused_status.execute(
+    rent_use_case_factory.set_tenancy_paused_status.execute(
```
mrosel: It looks like you've just blanket renamed these... But if feels like we've gone too far the other way. These don't directly relate to `rent`

elena-vi: will rename this to set_case_paused_status

app/jobs/application_job.rb
```diff
@@ -1,6 +1,6 @@
 class ApplicationJob < ActiveJob::Base
-  def income_use_case_factory
-    @income_use_case_factory ||= Hackney::Income::UseCaseFactory.new
+  def rent_use_case_factory
+    @rent_use_case_factory ||= Hackney::Rent::UseCaseFactory.new
   end
```
elena-vi: keep as rent

cob16:
One of the things here is that we do have a rent API in place. There may be duplicate functionality or a clash of naming (i.e the definition of the domain rent could be different) here. This could cause confusion with developers between the two.

PR #131, opened 2019-03-18
Update uh simulator

cob16:
- the simulator now matches `uhlive` instead of `uhdev`
- removed the local reference (this will speed build time and ensure we are using the correct pinned version on local test runs)


cob16:
As we now have primary keys in msql the helpers will have to turn on IDENTITY_INSERT mode to be able to set these fields

PR #133, opened 2019-03-21
Mr step 3 mob setup end to end

elena-vi:


lib/hackney/income/process_letter.rb
```diff
@@ -0,0 +1,22 @@
...
+      end
+
+      def execute(uuid:, user_id:)
+        html = Rails.cache.read(uuid)
+        pdf_obj = @pdf_generator.execute(html)
+
+        file_obj = pdf_obj.to_file("tmp/#{uuid}.pdf")
+
+        File.delete("tmp/#{uuid}.pdf")
+        Rails.cache.delete(uuid)
```
elena-vi: - [ ] maybe extract the following into a method?

lib/hackney/pdf/use_case_factory.rb
```diff
@@ -15,6 +15,13 @@ def get_preview
           leasehold_information_gateway: Hackney::ServiceCharge::UseCaseFactory.new.get_leasehold_information
         )
       end
+
+      # TODO: FIX AND REMOVE
+      # def generate_pdf
+      #   generator = Hackney::PDF::PDFGateway.new
+      #   html_preview = get_preview.html
+      #   generator.generate_pdf(html_preview).to_pdf
+      # end
```
elena-vi: - [ ] remove this because we've implemented this already?

spec/controllers/letters_controller_spec.rb
```diff
@@ -27,20 +27,49 @@
   end
 
   describe '#send_letter' do
+    # before { post :send_letter, params: { uuid: uuid, user_id: user_id } }
```
elena-vi: - [ ] remove

spec/controllers/letters_controller_spec.rb
```diff
@@ -27,20 +27,49 @@
...
+      it 'calls the usecase' do
+        expect_any_instance_of(Hackney::Income::ProcessLetter)
+          .to receive(:execute).with(uuid: uuid, user_id: user_id)
+
+        post :send_letter, params: { uuid: uuid, user_id: user_id }
+      end
+
+      # TODO: expect to send
+      # expect_any_instance_of(Hackney::Notification::SendManualPrecompiledLetter).to receive(:execute)
+    end
```
elena-vi: - [ ] Implement job to kick this off

spec/lib/hackney/pdf/generator_spec.rb
```diff
@@ -1,12 +1,25 @@
...
-      gateway.generate_pdf(html)
+      gateway.execute(html)
+    end
+
+    let(:source) { subject.source.instance_values['source'] }
+
+    it 'is a pdfkit' do
+      # byebug
+      # pdf_obj.to_file('magic_thing.pdf')
+      # File.open('magic_thing.pdf')
```
elena-vi: - [ ] remove hacky inline testing

Gemfile
```diff
@@ -60,6 +60,7 @@ group :development, :test do
   gem 'sqlite3'
   gem 'timecop'
   gem 'webmock'
+  gem 'pry'
```
elena-vi: it's about time 😆

app/jobs/hackney/income/jobs/send_letter_to_gov_notify_job.rb
```diff
@@ -8,29 +8,23 @@ def perform(document_id:)
...
-          document_metadata = get_metadata(document)
-
           letter_pdf = pdf_file_from_s3(document.filename)
 
           income_use_case_factory.send_precompiled_letter.execute(
-            user_id: document_metadata[:user_id],
-            payment_ref: document_metadata[:payment_ref],
-            unique_reference: unique_reference,
+            user_id: get_metadata(document)[:user_id],
+            payment_ref: get_metadata(document)[:payment_ref],
```
elena-vi: aren't you parsing the metadata twice here by calling them like that here?

Gemfile
```diff
@@ -55,6 +55,7 @@ group :development, :test do
   gem 'factory_bot'
   gem 'guard-rspec', require: false
   gem 'guard-rubocop', require: false
+  gem 'pry'
```
cob16: Do we still need this on production?

lib/hackney/cloud/adapter/aws_s3.rb
```diff
@@ -6,27 +6,42 @@ def initialize(encryption_client)
...
           # Add encrypted item to bucket
-          resp = client.put_object(
+          response = client.put_object(
             body: content,
             bucket: bucket_name,
-            key: new_filename
+            key: filename
           )
 
-          resp.successful? || raise('Cloud Storage Error!')
```
cob16: is it not necessary to check successful?

lib/hackney/cloud/adapter/aws_s3.rb
```diff
@@ -6,27 +6,42 @@ def initialize(encryption_client)
...
+          parse_response_context(response)
         end
 
-        def download(bucket_name, filename)
-          client.get_object(bucket: bucket_name, key: filename)
+        def download(bucket_name:, filename:)
+          # NOTE: MUST open and save a tmp file otherwise aws loses the encoding
+          temp_file_location = "tmp/#{filename}"
```
cob16: We should use rails temp file handling here https://ruby-doc.org/stdlib-1.9.3/libdoc/tempfile/rdoc/Tempfile.html

is this saving to a relative directory called tmp or the main /temp dir?

The difference is that /tmp is now a ramdisk on many Linux machines and will auto clean old files. where if it was relative we could fill the volume very quickly.

spec/hackney/income/jobs/save_and_send_letter_job_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  let(:enqueue_save_send) {
+    described_class.perform_now(bucket_name: bucket_name,
+                                filename: file_name,
+                                content: file.read,
+                                document_id: doc.id)
+  }
+
+  before {
+    expect_any_instance_of(Aws::S3::Encryption::Client).to receive(:put_object).and_return(AwsEncryptionClientDouble.new(nil).send(:put_object))
```
cob16: What happened here? I remember we were going to use doubles and stub constant https://github.com/LBHackney-IT/lbh-income-api/blob/4ba98bd5937dec9ba9c2d48f826b352d6aeeaf46/spec/hackney/income/jobs/send_green_in_arrears_msg_job_spec.rb#L13

Gemfile
mrosel: definitely not... that's why it's in `:development, :test`. :stuck_out_tongue_winking_eye:

lib/hackney/cloud/adapter/aws_s3.rb
mrosel: This is the relative directory to #{rails root}/tmp/
We struggled to get gov notify to accept anything but a `File` object. That's the only reason we write it out, then clean it up right after. I agree this would be a better solution overall.

cob16: So this will use up all the available space on the docker nodes then?

mrosel: it won't. Because `Tempfile` does not work, we have to use a `File` object. The solution was to treat it like a temp file here https://github.com/LBHackney-IT/lbh-income-api/pull/133/files#diff-a80eea6a88b3baa079b283b96eeea3feR29 and

mrosel: As an alternative we could move the temp logic into the download method and retun the `File` object instead of the file path

lib/hackney/notification/gov_notify_gateway.rb
```diff
@@ -14,7 +14,7 @@ def initialize(sms_sender_id:, api_key:, send_live_communications:, test_phone_n
       end
 
       def send_text_message(phone_number:, template_id:, reference:, variables:)
-        responce = @client.send_sms(
+        responce = client.send_sms(
```
alexdesi: nice change, can you meanwhile  fix the wrong spelling? 
respon**s**e

lib/hackney/notification/gov_notify_gateway.rb
```diff
@@ -25,7 +25,7 @@ def send_text_message(phone_number:, template_id:, reference:, variables:)
       end
 
       def send_email(recipient:, template_id:, reference:, variables:)
-        responce = @client.send_email(
+        responce = client.send_email(
```
alexdesi: as before

lib/hackney/pdf/preview.rb
```diff
@@ -14,16 +14,27 @@ def execute(payment_ref:, template_id:)
...
+            errors: preview_with_errors[:errors]
+          }
+        )
       end
 
       private
 
+      def save_to_cache(cache_key:, cache_value:)
+        Rails.cache.write(cache_key, cache_value, expires_in: 12.hours)
+        cache_value
```
alexdesi: We might  need not to return this, the caller already know the cache_value.

spec/controllers/letters_controller_spec.rb
```diff
@@ -27,19 +27,41 @@
   end
 
   describe '#send_letter' do
+    context 'when user "accepts" the preview' do
+      let(:user_id) { Faker::Number.number }
+      let(:uuid) { SecureRandom.uuid }
+
+      xit 'calls succefully' do
```
alexdesi: this sounds reasonable, do we want to implement it?

spec/controllers/letters_controller_spec.rb
```diff
@@ -49,15 +71,15 @@
       let(:missing_mandatory_data) { 222 }
 
       it 'no errors when only optional data is missing' do
```
alexdesi: context suggested here:
```
context  'when only optional data is missing do
  let(:missing_mandatory_data) { 222 }

it 'return success 200' do 
end
end

Comment (alexdesi) [reply spec/controllers/letters_controller_spec.rb]: ops it's out of scope, nevermind :)

File: spec/hackney/income/jobs/save_and_send_letter_job_spec.rb [hunk 17]
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  it 'uploads to clouds' do
+    enqueue_save_send
+    uploaded_doc = Hackney::Cloud::Document.find(doc.id)
+    expect(uploaded_doc.url).to eq 'blah.com'
```
alexdesi: nitpick: line before expectation

spec/hackney/income/jobs/save_and_send_letter_job_spec.rb
alexdesi: I think for similar think I've used 

```Hackney::Cloud::Adapter::Fake```
this is my current implementation (in the PR, to be merged)

```ruby
module Hackney
  module Cloud
    module Adapter
      class Fake
        @cloud_storage = {}

        def upload(bucket_name:, filename:, new_filename:)
          Fake.save(new_filename, File.read(filename))
        end

        def download(bucket_name, filename)
          Fake.read(filename)
        end

        def self.save(key, value)
          @cloud_storage[key] = value
        end

        def self.read(key)
          @cloud_storage[key]
        end
      end
    end
  end
end
```

spec/hackney/income/jobs/send_letter_to_gov_notify_job_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  class AwsResponse
+    def key; end
+
+    def body
+      StringIO.new
+    end
+  end
+end
```
alexdesi: this test looks very smart, but it took 5 min to understand what it does 😃 
Expectations are spread in the before, after and inside the test case, and it is not easy to see what we are testing (the `Act` fase).
Any idea to make life easier to who (like me) is more used to classic `Arrange Assert Act`?

spec/lib/hackney/cloud/adapter/aws_s3_spec.rb
```diff
@@ -1,51 +1,30 @@
...
+  let(:s3) { described_class.new Hackney::Cloud::EncryptionClient.new(ENV['CUSTOMER_MANAGED_KEY']).create }
+  let(:filename) { 'test_key.pdf' }
+  let(:file) { File.open('spec/test_files/test_pdf.pdf', 'rb') }
+  let(:content) { file.read }
 
-  let(:encryption_client_double) { double }
+  let(:bucket_name) { 'hackney-docs-development' }
 
-  let(:new_filename) { 'new_filename.txt' }
+  context 'when upload' do
```
alexdesi: this looks like a `describe '#upload`

spec/lib/hackney/cloud/adapter/aws_s3_spec.rb
```diff
@@ -1,51 +1,30 @@
...
   end
 
-  context 'when there are upload errors' do
-    let(:upload_response) { double(successful?: false) }
-
-    it 'raises an exception' do
-      allow(encryption_client_double).to receive(:put_object)
-        .with(body: content, bucket: 'my-bucket', key: new_filename)
-        .and_return(upload_response)
+  context 'when download' do
```
alexdesi: this looks like a `describe '#download`

app/jobs/hackney/income/jobs/send_letter_to_gov_notify_job.rb
```diff
@@ -0,0 +1,39 @@
...
+
+        def pdf_file_from_s3(filename)
+          temp_file_location =
+            Rails.configuration
+                 .cloud_adapter
+                 .download(bucket_name: HACKNEY_BUCKET_DOCS, filename: filename)
+          pdf = File.open(temp_file_location)
+          File.delete(temp_file_location)
+          pdf
+        end
```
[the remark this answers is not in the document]
mrosel: This has for gov notify - regardless it has been remedied with a temp object for GC to clean up

spec/support/stubs/aws_encryption_client_double.rb
```diff
@@ -0,0 +1,15 @@
...
+    context
+  end
+
+  def context
+    OpenStruct.new(
+      context: OpenStruct.new(
+        http_request: OpenStruct.new(headers: { 'x-amz-date' => Time.new(2002).to_s }, endpoint: 'blah.com')
+      )
+    )
+  end
```
alexdesi: do we need OpenStruct here? is suppose if you use it it's because hash are not enought.
What about just `Struct`? (I prefer to do not use OpenStruct because if you mispell an attribute it difficult to find out the error)

mrosel: The alternative is to use nested classes. However, all we care about here is behavior. Hence, openstruct. It emulates a class structure that aws uses in order to effectively double it.
We could implement a mirror of the aws class structure but IMO that's doesn't provide additional benefit.

spec/lib/hackney/cloud/adapter/aws_s3_spec.rb
mrosel: Ofcourse, this isn't a hard and fast rule but... 
following `http://www.betterspecs.org/` -  We aren't testing the method of the described class but an implementation of it, so `context` is used.

spec/controllers/letters_controller_spec.rb
mrosel: good catch! I think this was oversight

lib/hackney/pdf/preview.rb
mrosel: :+1:

lib/hackney/notification/gov_notify_gateway.rb
mrosel: prefer not to touch code that is unrelated to the changes and this is referenced elsewhere. I've tagged it for a future refactor (it's misspelled a lot)

lib/hackney/notification/gov_notify_gateway.rb
```diff
@@ -4,8 +4,8 @@ module Hackney
   module Notification
     class GovNotifyGateway
       def initialize(sms_sender_id:, api_key:, send_live_communications:, test_phone_number: nil, test_email_address: nil)
+        @api_key = api_key
         @sms_sender_id = sms_sender_id
-        @client = Notifications::Client.new(api_key)
         @send_live_communications = send_live_communications
 
         # TODO: do something with these
```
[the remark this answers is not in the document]
mrosel: in favor of memoization and the ability to mock. :smile:

lib/hackney/cloud/storage.rb
```diff
@@ -9,32 +9,34 @@ def initialize(storage_adapter, document_model)
...
         if new_doc.errors.empty?
-          Hackney::Cloud::Jobs::SaveToCloudJob.perform_later(bucket_name: HACKNEY_BUCKET_DOCS,
-                                                             filename: filename,
-                                                             new_filename: new_filename,
-                                                             model_document: document_model.name,
-                                                             uuid: uuid)
+          file.rewind
+          Hackney::Income::Jobs::SaveAndSendLetterJob.perform_later(
+            bucket_name: HACKNEY_BUCKET_DOCS,
+            content: file.read,
```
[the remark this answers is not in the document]
mrosel: I'm going to test this further because you are right... it feels troublesome

lib/hackney/cloud/adapter/aws_s3.rb
```diff
@@ -6,27 +6,42 @@ def initialize(encryption_client)
...
 
+        def parse_response_context(response)
+          http_request_context = response.context.http_request
+
+          uploaded_at = Time.parse(http_request_context.headers['x-amz-date'])
+
+          url = http_request_context.endpoint
+          { url: url, uploaded_at: uploaded_at }
+        end
+
```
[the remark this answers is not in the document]
mrosel: agreed. It reads easier to have the response on its own line.

lib/hackney/cloud/adapter/aws_s3.rb
mrosel: Got this to work with `Tempfile`. Please take a look at the implementation. It doesn't look as nice but It should have better behavior.

lib/hackney/cloud/adapter/aws_s3.rb
mrosel: the client handles this natively. It was uncovered whiled passing stubs to aws

spec/hackney/income/jobs/save_and_send_letter_job_spec.rb
mrosel: @alexdesi your implementation stubs it a level higher. What we've implemented here is closer to a mock for Aws::S3::Encryption::Client. IMO this makes the tests less fragile

spec/hackney/income/jobs/send_letter_to_gov_notify_job_spec.rb
mrosel: :thinking: Could you explain what you mean? 
It's meant to be `Arrange, Act, Assert` - but you have it the other way
IMO, `expect_any_instance_of` is confustion because it defers the assertion

arrange:
`let(:document_id) ...`

Act: 
` described_class.perform_now ... `

Assert:
`expect_any_instance_of(Hackney: ... `

PR #134, opened 2019-03-26
Add new endpoint to download a letter

alexdesi:
This add a new endpoint to download a letter previously stored by UUID.
Note:
It makes *partial use* of Clearn Architecture (it does not use the gatway for the document table,  it is something that I can add, but I noticed that we do not always use gatway, sometimes we just use active record).

Trello: https://trello.com/c/2MXgXVlS/164-view-saved-letter


app/controllers/letters_controller.rb
```diff
@@ -0,0 +1,9 @@
+class LettersController < ApplicationController
+  LETTER_FILE_NAME = 'letter.pdf'.freeze
```
elena-vi: maybe we could call this letter as something a little more _meaningful_, for example maybe having the file name be the payment_ref and letter template name? 

I'm just thinking of the user experience here and I'm assuming if a user is downloading several letters at once it would be confusing for them to download a bunch of letters that are just called `letter.pdf`, `letter(1).pdf`, `letter(2).pdf`....

lib/hackney/cloud/storage.rb
```diff
@@ -33,8 +33,20 @@ def save(filename)
         { errors: new_doc.errors.full_messages }
       end
 
+      def read_document(uuid)
+        doc = document_model.find_by(uuid: uuid)
```
elena-vi: why are we finding by `uuid` here? I thought we settled on fetching by `id`s

lib/hackney/letter/download_use_case.rb
```diff
@@ -0,0 +1,13 @@
+module Hackney
+  module Letter
+    class DownloadUseCase
+      def initialize(cloud_storage)
+        @cloud_storage = cloud_storage
+      end
+
+      def execute(uuid:)
+        @cloud_storage.read_document(uuid)
```
elena-vi: same here on fetching by `uuid` also, it seems that your cloud storage is also doing the fetching from our database. I think it might make more sense to fetch the document record from our database here and then just pass in the document filename and bucket name into the `cloud_storage` to simply do the fetching from aws

spec/hackney/cloud/jobs/save_to_cloud_job_spec.rb
```diff
@@ -5,18 +5,19 @@
 
   let(:uuid) { SecureRandom.uuid }
 
+  let(:filename) { './spec/lib/hackney/cloud/adapter/upload_test.txt' }
```
elena-vi: I'm confused why this `filename` is a path to a file..

config/routes.rb
```diff
@@ -13,5 +13,7 @@
     get '/messages/get_templates', to: 'messages#get_templates'
     get '/pdf/get_templates', to: 'pdf#get_templates'
     post '/pdf/send_letter', to: 'pdf#send_letter'
+
+    get '/letters/:uuid', to: 'letters#download'
```
mrosel: This doesn't match the discussed AC. Must change this to reflect the primary key and not an attribute. If we want to have a uuid, then we should pass this in a param `?uuid=`

app/controllers/letters_controller.rb
alexdesi: I like the idea, but not superhigh priority.
I created a trello for this: https://trello.com/c/rapsF79W/183-use-meaningful-letter-filename,
thanks!

lib/hackney/cloud/storage.rb
alexdesi: yes, np

alexdesi: even thought I think download the doc by uuid is important, since we store it in the filename on S3. Probably we should support both.

elena-vi: I agree, you download from S3 by uuid, but we should probably be fetching from our db by id

app/controllers/letters_controller.rb
mrosel: Suggest we change this up to reflect something meaningful.
I know I get frustrated with other sites when all the downloads conflict. 

IMO, the filename with the uuid would be better than the same name for every file.

spec/hackney/cloud/jobs/save_to_cloud_job_spec.rb
mrosel: we have a tested file that I would suggest reusing `spec/test_files/test_pdf.pdf`

lib/hackney/letter/download_use_case.rb
alexdesi: I did it for consistency, 
`Storage#save` creates an entry in `Cloud::Document` and saves in the cloud the doc.
so
`Storage#read_document` read the `Cloud::Document` and read from the Cloud as well.

app/controllers/letters_controller.rb
alexdesi: I agree, but as discussed, not really high priority. A new trello card has been created for this.

spec/hackney/cloud/jobs/save_to_cloud_job_spec.rb
alexdesi: Mark that file is quite big (27KB), and I do not need to handle PDF. the upload_test.txt is just few bytes, I think is is more convenent to use it (we do not care the type of files for these tests)

alexdesi: @elena-vi I agree, for sure when we integrate this branch with that one you and mark are working, we'll unify that name.

alexdesi:
@mrosel if you and @elena-vi have completed your branch, let's merge this, and create the ?pay_ref=xxx for a different PR, since it is not high priority.

PR #139, opened 2019-04-01
Add makefile cmd -> update simulator

cob16:
Simplify troubleshooting you local build with this one simple cmd! :smile:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,14 @@ bundle:
 .PHONY: setup
 setup: docker-build bundle
 
+.PHONY: update-simulator
+update-simulator:
+	docker kill $(docker ps --filter "name=universal_housing_simulator" -q) || :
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -14,6 +14,14 @@ bundle:
 .PHONY: setup
 setup: docker-build bundle
 
+.PHONY: update-simulator
+update-simulator:
```
cybojenix: prefer leveraging docker-compose for this.

```sh
docker-compose rm -vs universal_housing_simulator
$(MAKE) ecr-login
docker-compose pull universal_housing_simulator
```

You can also do this:
```sh
$(MAKE) ecr-login
docker-compose up -d --force-recreate -V universal_housing_simulator
```

which will create a new container, clear the volumes, and swap in the new service without taking anything else down

PR #140, opened 2019-04-01
164 view saved letter integration

alexdesi:
This branch has been created to integrate branch 164 (download document) with master.


lib/hackney/cloud/adapter/aws_s3.rb
```diff
@@ -17,6 +17,11 @@ def upload(bucket_name:, binary_letter_content:, filename:)
           parse_response_context(response)
         end
 
+        # def download(bucket_name, filename)
```
mrosel: Could you annotate these comments with a comment, or just remove?

lib/hackney/cloud/storage.rb
```diff
@@ -33,6 +33,16 @@ def save(letter_html:, uuid:, filename:, metadata:)
         { errors: new_doc.errors.full_messages }
       end
 
+      def read_document(id)
+        doc = document_model.find_by(id: id)
```
mrosel: suggest using `find`

```suggestion
        doc = document_model.find(id)
```

app/controllers/documents_controller.rb
```diff
@@ -0,0 +1,9 @@
+class DocumentsController < ApplicationController
+  LETTER_FILE_NAME = 'letter.pdf'.freeze
+
+  def download
+    response = letter_use_case_factory.download.execute(id: params.fetch(:id))
+
+    send_data response[:content], filename: LETTER_FILE_NAME
```
mrosel: TIL: :tada:

elena-vi:
I'm confused.. how is this different from this https://github.com/LBHackney-IT/lbh-income-api/pull/134 ?

lib/hackney/cloud/storage.rb
alexdesi: actually those are different, your suggestion raises an exception in case the ID is not found, but I'll accept your suggestion  and remove the following `raise 'File does not exist!' if doc.nil?`

PR #146, opened 2019-04-03
Potential Bug Fix for the plain white PDF bug.

alexdesi:
This is a potential bugfix (discussed with @mrosel) for the plain white PDF generated on Staging env.

### DIFF
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -4,6 +4,6 @@ class DocumentsController < ApplicationController
   def download
     response = letter_use_case_factory.download.execute(id: params.fetch(:id))
 
-    send_data response[:content], filename: LETTER_FILE_NAME
+    send_file response[:filepath], type: 'application/pdf', filename: LETTER_FILE_NAME
   end
 end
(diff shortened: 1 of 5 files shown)


app/controllers/documents_controller.rb
```diff
@@ -4,6 +4,6 @@ class DocumentsController < ApplicationController
   def download
     response = letter_use_case_factory.download.execute(id: params.fetch(:id))
 
-    send_data response[:content], filename: LETTER_FILE_NAME
+    send_file response[:filepath], type: 'application/pdf', filename: LETTER_FILE_NAME
```
mrosel: Aren't we storing `'application/pdf'` in the metadata of the `Document` model? Might be easier to use that instead of hard coding the value here.

spec/controllers/documents_controller_spec.rb
```diff
@@ -16,7 +16,7 @@
 
       it 'download the document' do
         expect(download_use_case).to receive(:execute).with(id: id)
-        expect_any_instance_of(described_class).to receive(:send_data).with('Hello Hackney', filename: 'letter.pdf')
+        expect_any_instance_of(described_class).to receive(:send_file).with('example.pdf', type: 'application/pdf', filename: 'letter.pdf')
```
mrosel: Suggest that we use `allow` here to test the actual rendering 
`allow(described_class).to receive(:render)` should do the trick

alexdesi:
@mark I agree with your comments, but at the moment we want to really check if the `sent_file` fixes the bug.

PR #147, opened 2019-04-05
enable enable_x_sendfile_header to download PDF

alexdesi:
The PDF issue could be caused by a server configuration regarding sending file to the client.


alexdesi:
Not sure about the validity of the solution.

PR #148, opened 2019-04-08
Polling for letter state changes

mrosel:
Polls GovNotify every n minutes for status changes of qualifying documents.

### DIFF
--- a/schedule.yml
+++ b/schedule.yml
@@ -12,3 +12,7 @@
   green_in_arrears_messages:
     class: GreenInArrearsMessages
     cron: '0 0 9 * * 5' # 9am utc every friday
+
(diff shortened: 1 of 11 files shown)


schedule.yml
```diff
@@ -12,3 +12,7 @@
   green_in_arrears_messages:
     class: GreenInArrearsMessages
     cron: '0 0 9 * * 5' # 9am utc every friday
+
+  request_all_precompiled_letter_states:
+    class: RequestAllPrecompiledLetterStates
+    every: 5m
```
mrosel: :thinking: @elena-vi / @alexdesi

elena-vi: I think we might just need to make this once or maybe twice a day

PR #152, opened 2019-04-16
Remove property data stub

alexdesi:
This PR make use of real data from UH DB to populate the letters.

Trello:
https://trello.com/c/uRPsaXWj/181-remove-property-data-stub-and-retrieve-data-from-uh-db


lib/hackney/service_charge/gateway/service_charge_gateway.rb
```diff
@@ -23,6 +23,8 @@ def initialize(host:, api_key:)
         end
 
         def get_cases_by_refs(refs)
+          # TODO: retrieve the fields from UH DB
+          # using adapters
```
elena-vi: I find these comments confusing, also surely implementing the UH solution would mean that everything after this method would have been deleted as we're replacing all uses of `fake_get_cases_by_refs` by the call to UH

spec/controllers/letters_controller_spec.rb
```diff
@@ -6,23 +6,39 @@
   let(:template_name) { 'Letter 1 template' }
   let(:preview_html) { "<p>#{Faker::HitchhikersGuideToTheGalaxy.quote}</p>" }
 
+  class StubGetTemplates
```
elena-vi: is this used anywhere?

app/controllers/letters_controller.rb
```diff
@@ -10,7 +10,7 @@ def create
       payment_ref: params.fetch(:payment_ref),
       template_id: params.fetch(:template_id)
     )
-  rescue Hackney::ServiceCharge::Exceptions::ServiceChargeException
+  rescue ArgumentError
```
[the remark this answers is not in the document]
mrosel: also, this looks out of scope for this pr. I'm curious how this rescue affects the UH queries?

lib/hackney/income/universal_housing_leasehold_gateway.rb
```diff
@@ -0,0 +1,65 @@
...
+          property_address_1: property_res[:post_preamble]
+        }
+      end
+
+      private
+
+      def get_prop_ref(payment_ref)
+        u_letsvoids = database[:u_letsvoids].first(payment_ref: payment_ref)
+
+        raise ArgumentError, 'payment_ref does not exist!' unless u_letsvoids.present?
```
mrosel: Is this right? We're raising `payment_ref` but checking for `u_letsvoids`

spec/support/universal_housing_helper.rb
```diff
@@ -1,9 +1,11 @@
 module UniversalHousingHelper
-  def create_uh_tenancy_agreement(tenancy_ref:, current_balance: 0.0, property_ref: '', terminated: false, tenure_type: 'SEC', high_action: '111')
+  # rubocop:disable Metrics/ParameterLists
```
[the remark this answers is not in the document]
mrosel: this isn't bothering me as much because it's supporting legacy code. I would definitely prefer fixture like behavior here but that feels out of scope.

lib/hackney/income/universal_housing_leasehold_gateway.rb
mrosel:
```suggestion
```

lib/hackney/income/universal_housing_leasehold_gateway.rb
```diff
@@ -0,0 +1,65 @@
...
+          correspondence_address_5: corr_postcode_res[:aline4],
+          correspondence_address_6: househ_res[:corr_postcode],
+          property_address_1: property_res[:post_preamble]
+        }
+      end
+
+      private
+
+      def get_prop_ref(payment_ref)
+        u_letsvoids = database[:u_letsvoids].first(payment_ref: payment_ref)
```
mrosel:
```suggestion
        u_letsvoids = database[:u_letsvoids].first!(payment_ref: payment_ref)
```

lib/hackney/income/universal_housing_leasehold_gateway.rb
```diff
@@ -0,0 +1,65 @@
+module Hackney
+  module Income
+    class UniversalHousingLeaseholdGateway
+      def get_leasehold_info(payment_ref:)
+        prop_ref = get_prop_ref(payment_ref)
+
+        tenagree_res = tenagree.first(prop_ref: prop_ref)
+        rent_res = rent.first(prop_ref: prop_ref)
+        property_res = property.first(prop_ref: prop_ref)
+        househ_res = househ.first(house_ref: tenagree_res[:house_ref])
```
mrosel: I see cascading issues with this since this relies on `tenagree_res` but `tenagree_res` isn't guaranteed to be found. A null value either here or in the subsequent queries are likely to break.

spec/lib/hackney/income/universal_housing_leasehold_gateway_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+          aline2: 'Address line 2',
+          aline3: 'Address line 3',
+          aline4: 'Address line 4',
+          post_code: Faker::Address.postcode
+        }
+      }
+
+      # let(:corr_postcode) { Faker::Address.postcode }
+      let(:post_code) { Faker::Address.postcode }
+      let(:prop_ref) { '123' }
```
mrosel: Must use Faker for triangulation.

spec/lib/hackney/income/universal_housing_leasehold_gateway_spec.rb
```diff
@@ -0,0 +1,85 @@
+require 'rails_helper'
+
+describe Hackney::Income::UniversalHousingLeaseholdGateway, universal: true do
```
mrosel: like presented in my comments above https://github.com/LBHackney-IT/lbh-income-api/pull/152/files#r276146614
These tests would benefit from triangulation - we know that UH is likely to not be consistent but this is only testing the happy path

app/controllers/letters_controller.rb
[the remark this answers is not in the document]
alexdesi: good points,
@elena-vi I removed that because it is used for API calls, the message of that exception is:
`[Service Charge API error: Received #{response&.code} response] #{super}` 
that is not appropriate of UH access
@mrosel yep, the `ArgumentError` might be raised from different places, I'd suggest to create a new exception for UH Db, that we can use in case of missing reference.
We could use the existing namespace `Hackney::ServiceCharge::Exceptions`
or create a new one
`Hackney::UHDatabase::Exceptions`

lib/hackney/income/universal_housing_leasehold_gateway.rb
```diff
@@ -0,0 +1,65 @@
+module Hackney
+  module Income
+    class UniversalHousingLeaseholdGateway
+      def get_leasehold_info(payment_ref:)
+        prop_ref = get_prop_ref(payment_ref)
+
+        tenagree_res = tenagree.first(prop_ref: prop_ref)
```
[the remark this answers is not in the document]
alexdesi: you are right, that's the point 😄 
from my understanding we should use DB naming convention in the gateway, but just here!
and do not let leak those *uniqueness* in other internal objects/layers.
so the output hash should be simple and without weird name.

ps. if you you meant at `..._res` that's my convention to say `result of the query`, I'm happy to change it to something more meaningful.

lib/hackney/income/universal_housing_leasehold_gateway.rb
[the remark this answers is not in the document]
alexdesi: I don't know, what I know is that I can extract the `prop_ref` from that table: `start_date |  end_date | payment_ref | account_ref | prop_ref`
Anyway I agree with you that we can have a better name for that variable, I'll ask Nick or Elaine.

lib/hackney/income/universal_housing_leasehold_gateway.rb
alexdesi: yep, the previous line is:
`u_letsvoids = database[:u_letsvoids].first(payment_ref: payment_ref)`
(I'll try to find a better name for this variable)

lib/hackney/income/universal_housing_leasehold_gateway.rb
```diff
@@ -0,0 +1,65 @@
...
+        rent_res = rent.first(prop_ref: prop_ref)
```
mrosel: Please note `first` with `Sequel` does not behave the same as `ActiveRecord` and this isn't guaranteed to return the correct rent account. If you need ordering, you will have to add it here. Please double check as I'm not familiar with the relationship with `rent`

lib/hackney/service_charge/gateway/service_charge_gateway.rb
alexdesi: I agree, I forgot to remove those.

spec/support/universal_housing_helper.rb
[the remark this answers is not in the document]
alexdesi: Becase of 🤖 `rubocop` suggest a max of 7 params for the methods,
we got 8 here: `tenancy_ref:, current_balance: 0.0, prop_ref: '', terminated: false, tenure_type: 'SEC', high_action: '111', u_saff_rentacc: '', house_ref: ''`
AFAIK those are all used by some tests, so we would need to refactor this methods in multiple ones.
I'd put it as low priority at the moment.

lib/hackney/income/universal_housing_leasehold_gateway.rb
alexdesi: you're right, we need a strategy then, I can see 2 choises:
1. raise an error and do not create letter previous?
2. poputate with empty values whenever we cannot retrieve the data
what is better for the user? maybe the second one

lib/hackney/income/universal_housing_leasehold_gateway.rb
alexdesi: I removed that ugly `u_letsvoids` .
Regarding `first!` I like the idea, but in case of error it raises 
`Sequel::NoMatchingRow: Sequel::NoMatchingRow`
that we need to catch in the controller, and we do not want to couple the controllers with that exceptions

lib/hackney/income/universal_housing_leasehold_gateway.rb
alexdesi: it's more readable now

lib/hackney/income/universal_housing_leasehold_gateway.rb
alexdesi: I added a couple of joins to improve in this point, have a look to the updates.

spec/lib/hackney/pdf/preview_spec.rb
```diff
@@ -9,7 +9,7 @@
   end
 
   let(:get_templates_gateway) { instance_double(Hackney::PDF::GetTemplates) }
-  let(:leasehold_information_gateway) { instance_double(Hackney::ServiceCharge::GetLeaseholdInformation) }
+  let(:leasehold_information_gateway) { instance_double(Hackney::Income::UniversalHousingLeaseholdGateway) }
```
cybojenix: The `test_letter_params` will need updating to reflect the new response structure

spec/lib/hackney/income/universal_housing_leasehold_gateway_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+          original_lease_date: sc_leasedate,
+          lessee_full_name: house_desc,
+          date_of_current_purchase_assignment: cot
+        }.merge(correspondence_address)
+         .merge(property_address))
+      end
+    end
+  end
+
+  def correspondence_address
```
cybojenix: This doesn't align with what `Hackney::ServiceCharge::Letter` expects.

Suggest an integration test be added.

Also think that `Letter` is under the wrong namespace now

lib/hackney/income/universal_housing_leasehold_gateway.rb
```diff
@@ -0,0 +1,67 @@
+module Hackney
+  module Income
+    class UniversalHousingLeaseholdGateway
+      def get_leasehold_info(payment_ref:)
+        res = tenagree
+              .where(u_saff_rentacc: payment_ref)
+              .join(rent, prop_ref: :prop_ref)
+              .join(househ, prop_ref: :prop_ref)
```
cybojenix: seems the database doesn't always have `prop_ref` on the house hold table. This should be raised with someone to find out what this means and if we need to care

lib/hackney/income/universal_housing_leasehold_gateway.rb
```diff
@@ -0,0 +1,67 @@
...
+          correspondence_address_5: corr_postcode_res[:aline4],
+          correspondence_address_6: res[:corr_postcode],
+
+          property_address: property_res[:address1] + ', ' + property_res[:post_code]
+        }
+      end
+
+      private
+
+      def tenagree
```
cybojenix: These properties are within our namespace, so we can give them their proper names at this point.

`tenagree` -> `tenancy_agreement`
`househ` -> `household`

spec/lib/hackney/income/universal_housing_leasehold_gateway_spec.rb
cybojenix: very much agree with this, to the point that it'd be worth writing a fuzzer to replace the existing factories. We've already found some inconsistencies with addresses and property references

spec/controllers/letters_controller_spec.rb
```diff
@@ -6,23 +6,39 @@
...
 
       get :get_templates
 
-      expect(response.body).to eq(
-        {
-          path: template_path,
-          id: template_id,
-          name: template_name
-        }.to_json
-      )
```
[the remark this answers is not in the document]
alexdesi: I think you are right, that's a kind of integration test. I pruned too much :)

lib/hackney/income/universal_housing_leasehold_gateway.rb
alexdesi: I'll discuss with Soraya, about this point.

spec/lib/hackney/income/universal_housing_leasehold_gateway_spec.rb
alexdesi: I added unhappy path on the test, to cover possible missing data con `property` and `postcode` table

lib/hackney/income/universal_housing_leasehold_gateway.rb
alexdesi: for now let's merge considering just the prop_ref not empty.
I've explicity coded this condition: 
```
res = tenagree
              .where(u_saff_rentacc: payment_ref)
              .exclude(Sequel.trim(Sequel.qualify(:tenagree, :prop_ref)) => '')
```

lib/hackney/income/universal_housing_leasehold_gateway.rb
```diff
@@ -0,0 +1,74 @@
...
+
+      def property
+        @property ||= database[:property]
+      end
+
+      def household
+        @household ||= database[:househ]
+      end
+
+      def get_short_name(full_name)
```
mrosel: Might as well just remove this since we're using full name

spec/lib/hackney/pdf/preview_spec.rb
```diff
@@ -22,21 +22,23 @@
...
       correspondence_address3: 'London',
+      correspondence_address4: 'London',
+      correspondence_address5: 'England',
       correspondence_postcode: 'SE1 0SW',
-      lessee_short_name: 'Philip',
       property_address: '1 Hillman St, London, E8 1DY',
-      arrears_letter_1_date: '20th Feb 2019',
-      total_collectable_arrears_balance: '3506.90'
+      arrears_letter_1_date: '20th Feb 2019', # TODO: this will be necessary for letter2
+      total_collectable_arrears_balance: '3506.90' # TODO: change this to balance
```
mrosel: Are these still todo?

spec/lib/hackney/service_charge/gateway/service_charge_gateway_spec.rb
```diff
@@ -1,89 +0,0 @@
...
-          ] }.to_json
-        )
-      end
-
-      it 'recognises international correspondence address' do
-        expect(subject.first[:international]).to eq(false)
-      end
-    end
-
-    context 'when the case has a ref and international correspondence address' do
```
mrosel: This was part of the original specification and to my knowledge is still valid. However, I don't see that it has been reimplemented anywhere.

spec/lib/hackney/pdf/preview_spec.rb
alexdesi: I can removed them, since I created relevant trello cards for them.

spec/lib/hackney/service_charge/gateway/service_charge_gateway_spec.rb
alexdesi: I put international back, and updated the test.

alexdesi:
@mrosel I've fixed last things and 
just merged master into this feature branch (to avoid surprises),
I'll merge as soon you accept it.

PR #154, opened 2019-04-25
Mr send letter action diary

mrosel:
1. fetches the tenancy_ref from uh when given the payment_ref
2. uses the tenancy_ref and posts a message to the UH action diary with a note.


lib/hackney/income/universal_housing_leasehold_gateway.rb
```diff
@@ -1,6 +1,15 @@
 module Hackney
   module Income
     class UniversalHousingLeaseholdGateway
+      def map_tenancy_ref_to_payment_ref(payment_ref:)
```
alexdesi: Being this the gatway for a DB it's better to call it something like:
`get_tenancy_ref` or 
`get_tenancy_ref_by` (so we can `get_tenancy_ref_by(payment_ref: 'xxxx')`

lib/hackney/income/universal_housing_leasehold_gateway.rb
```diff
@@ -1,6 +1,15 @@
...
+        res = tenancy_agreement
+              .where(u_saff_rentacc: payment_ref)
+              .first
+
+        raise TenancyNotFoundError unless res.present?
+        { tenancy_ref: res[:tag_ref] }
```
alexdesi: just to check, is `tag_ref` the `tenancy_ref`? (I had no idea)

lib/hackney/notification/send_manual_precompiled_letter.rb
```diff
@@ -1,13 +1,28 @@
...
+          notification_gateway.send_precompiled_letter(
+            unique_reference: unique_reference,
+            letter_pdf: letter_pdf
+          )
+
+        # FIXME: this must be in a background job => UH is unreliable
+        # TODO: create job to accept exact same args as add_action_diary_usecase
+
+        const = template_id.split(' ').join('_').upcase
+        action_code = "Hackney::Tenancy::ActionCodes::#{const}".constantize
```
alexdesi: what is const?
it looks like a class name, it is used only to get action_code, we can put it in a private method
```
def action_code
  ...
end
```

lib/hackney/service_charge/use_case_factory.rb
```diff
@@ -6,6 +6,12 @@ def get_leasehold_information
           service_charge_gateway: Hackney::Income::UniversalHousingLeaseholdGateway.new
         )
       end
+
+      def map_tenancy_information
```
alexdesi: this MapTenancy does not do much, since we are already using
`UniversalHousingLeaseholdGateway` here, you could simplify  and access to it got get the data you need.

lib/hackney/notification/send_manual_precompiled_letter.rb
```diff
@@ -1,13 +1,28 @@
...
-          unique_reference: unique_reference,
-          letter_pdf: letter_pdf
+      def execute(user_id: nil, payment_ref: nil, template_id:, unique_reference:, letter_pdf:)
+        send_letter_response =
+          notification_gateway.send_precompiled_letter(
+            unique_reference: unique_reference,
+            letter_pdf: letter_pdf
+          )
+
+        # FIXME: this must be in a background job => UH is unreliable
```
alexdesi: Do you want to create Trello for this on the backlog, so we do not forget?

lib/hackney/income/universal_housing_leasehold_gateway.rb
mrosel: yeah, check out - https://github.com/LBHackney-IT/lbh-income-api/blob/master/lib/hackney/income/tenancy_prioritiser/universal_housing_criteria.rb#L19

lib/hackney/income/universal_housing_leasehold_gateway.rb
mrosel: +1 for better naming

lib/hackney/service_charge/use_case_factory.rb
```diff
@@ -6,6 +6,10 @@ def get_leasehold_information
...
```
alexdesi: remove this

PR #156, opened 2019-04-26
Make preamble optional

alexdesi:
The preamble, that usually is the house name, should be optional (and most of the time is empty)

### DIFF
--- a/app/models/hackney/service_charge/letter.rb
+++ b/app/models/hackney/service_charge/letter.rb
@@ -2,7 +2,7 @@ module Hackney
   module ServiceCharge
     class Letter
       MANDATORY_LETTER_FIELDS = %i[payment_ref lessee_full_name
-                                   correspondence_address1 correspondence_address2
+                                   correspondence_address2 correspondence_address3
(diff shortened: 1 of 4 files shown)


app/models/hackney/service_charge/letter.rb
```diff
@@ -18,12 +18,12 @@ def initialize(params)
         validated_params = validate_mandatory_fields(params)
 
         @tenancy_ref = validated_params[:tenancy_ref]
-        @correspondence_address1 = validated_params[:correspondence_address1]
-        @correspondence_address2 = validated_params[:correspondence_address2]
-        @correspondence_address3 = validated_params[:correspondence_address3]
-        @correspondence_address4 = validated_params[:correspondence_address4]
-        @correspondence_address5 = validated_params[:correspondence_address5]
-        @correspondence_postcode = validated_params[:correspondence_postcode]
+        @correspondence_address1 = validated_params[:correspondence_address1] # corr_preamble ( the flat number/house Name)
```
mrosel: :100: the comments are actually extremely helpful here!

PR #162, opened 2019-04-29
Update letter_1_in_arrears_LH.erb

Soraya300:
Here are my changes, please modify.

I need you to:
1. Amend the following text in this ruby file. 
https://github.com/LBHackney-IT/lbh-income-api/blob/master/lib/hackney/pdf/templates/layouts/sender_address.erb
 
2. "Your ref" to "Payment Reference:"
and Our Ref for Letter 1 in arrears LH should read Our ref: SCARR1 (it is different for each template)

3. Remove space ("strip") after shortname.

4. Remove centre on green box and add to balance only. 

5. Make sure green box doesn't go over payment reference.

### DIFF
--- a/lib/hackney/pdf/templates/letter_1_in_arrears_LH.erb
+++ b/lib/hackney/pdf/templates/letter_1_in_arrears_LH.erb
@@ -15,12 +15,12 @@
   Dear <%= @letter.lessee_short_name %>,
 
   <p>
-    <strong>Leaserholder of the property:</strong> <br>
+    <strong>Property:</strong> <br>
     <%= @letter.property_address %>
   </p>
 
   <div class="green_box">
-    Our records show that you still have the outstanding amount on your rent & service charge account:
+    Our records show that you still have the outstanding amount on your service charge account:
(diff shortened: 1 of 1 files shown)


lib/hackney/pdf/templates/letter_1_in_arrears_LH.erb
```diff
@@ -42,21 +42,14 @@
...
     If you are experiencing financial difficulties, please contact your local Citizen Advice Bureau
-    or contact us to discuss any assistance we may be able to offer.
-  </p>
-  <p>
-    Please find enclosed further information on methods of payment.
-  </p>
-  <p>
-    If you have already made a payment or entered into an arrangement to pay the balance,
-    please disregard this letter.
+    or contact us to discuss any assistance we may be able to offer. Please find enclosed further information on methods of payment. If you have already made a payment or entered into an arrangement to pay the balance, please disregard this letter.
```
alexdesi: I think some of the paragraphs help for the readability.
Sure we want to remove all the `<p>`?

PR #166, opened 2019-05-01
Get action codes for Manual SMSs based on the template name

cybojenix:
Originally I planned to do this using template IDs, however as there are multiple templates that can be sent, I've opted for a looser approach that will work so long as the current structure is sent.

A fallback happens to send the previous Manual SMS action code (which is `GMS`) when the template name can't be identified.

### DIFF
--- a/lib/hackney/notification/manual_action_code.rb
+++ b/lib/hackney/notification/manual_action_code.rb
@@ -1,4 +1,3 @@
-
 module Hackney
   module Notification
     class ManualActionCode
@@ -23,4 +22,3 @@ module Hackney
     end
   end
 end
-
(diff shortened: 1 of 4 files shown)


lib/hackney/notification/manual_action_code.rb
```diff
@@ -0,0 +1,24 @@
+module Hackney
+  module Notification
+    class ManualActionCode
+      class << self
+        def get_by_sms_template_name(template_name:)
+          if template_name.start_with?('Green')
```
mrosel: SUGGEST not comparing with strings. Use symbols / enums instead.

cybojenix: reason for using a string is there are multiple templates that follow this structure.
I can change to using an enum, but it will tie adding a template to deploying code out.

at this stage, I'd rather lean on saying "follow this naming pattern you already have", as we want to avoid needing technical involvement for the next few months

PR #168, opened 2019-05-01
Add some rake tasks for BAU requests

cybojenix:
These are based on scripts posted into Slack.

Moving them to rake tasks to make them easier to run on the servers.

### DIFF
--- a/lib/tasks/bau/case_totals.rake
+++ b/lib/tasks/bau/case_totals.rake
@@ -1,0 +1,24 @@
+require 'json'
+
+namespace :bau do
+
+  desc 'generate case totals for each user'
+  task :case_totals, [:tenancy_ref, :user_id] do
+    user_cases = []
+    users = Hackney::Income::Models::User.all
+    users.each do |user|
+      cases = Hackney::Income::Models::CasePriority.where(assigned_user_id: user[:user_id])
+      green = cases.where(priority_band: :green)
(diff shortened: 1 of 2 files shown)


lib/tasks/bau/case_totals.rake
```diff
@@ -0,0 +1,23 @@
...
+      green = cases.where(priority_band: :green)
+      amber = cases.where(priority_band: :amber)
+      red = cases.where(priority_band: :red)
+      user_cases << {
+        name: user.name,
+        total: cases.length,
+        green: green.length,
+        amber: amber.length,
+        red: red.length
+      }
```
alexdesi: `length` is very inefficient, please use `count`!
for more: https://mensfeld.pl/2014/09/activerecord-count-vs-length-vs-size-and-what-will-happen-if-you-use-it-the-way-you-shouldnt/

lib/tasks/bau/case_totals.rake
```diff
@@ -0,0 +1,23 @@
+require 'json'
+
+namespace :bau do
+  desc 'generate case totals for each user'
+  task :case_totals do
+    user_cases = []
+    users = Hackney::Income::Models::User.all
```
alexdesi: This way we load the entire user table, and this could use whole memory, there are other ways to iterate throug the entire collection.

PR #169, opened 2019-05-01
add payment_ref filter to filter the documents

alexdesi:
It adds payment_ref filter to filter the documents

trello: https://trello.com/c/PbFUZGqU/203-add-payment-reference-url-to-document-view-page

### DIFF
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -7,7 +7,7 @@ class DocumentsController < ApplicationController
   end
 
   def index
-    render json: letter_use_case_factory.get_all_documents.execute
+    render json: letter_use_case_factory.get_all_documents.execute(payment_ref: params.fetch(:payment_ref, nil))
   end
 
   private
--- a/lib/hackney/letter/all_documents_use_case.rb
+++ b/lib/hackney/letter/all_documents_use_case.rb
@@ -5,8 +5,9 @@ module Hackney
         @cloud_storage = cloud_storage
       end
 
-      def execute
-        @cloud_storage.all_documents.each do |doc|
+      def execute(payment_ref: nil)
+        @cloud_storage.all_documents(payment_ref: payment_ref)
+                      .each do |doc|
           next unless doc.metadata
           metadata = JSON.parse(doc.metadata).deep_symbolize_keys
           if metadata[:user_id]
--- a/spec/lib/hackney/cloud/storage_spec.rb
+++ b/spec/lib/hackney/cloud/storage_spec.rb
@@ -12,8 +12,12 @@ describe Hackney::Cloud::Storage, type: :model do
     end
   end
 
-  it 'retrieves all documents' do
-    storage.all_documents
+  describe 'retrieve all document' do
+    it 'retrieves all documents' do
+      expect(Hackney::Cloud::Document).to receive(:all)
+
+      storage.all_documents
+    end
   end
 
   describe '#save' do
--- a/spec/lib/hackney/cloud/storage_spec.rb
+++ b/spec/lib/hackney/cloud/storage_spec.rb
@@ -18,6 +18,27 @@ describe Hackney::Cloud::Storage, type: :model do
 
       storage.all_documents
     end
+
+    context 'when payment_ref param is used' do
+      let(:payment_ref) { Faker::Number.number(10) }
+
+      before do
+        Hackney::Cloud::Document.create(filename: '-', uuid: SecureRandom.uuid, extension: '-', mime_type: '-',
+                                        metadata: { payment_ref: payment_ref }.to_json)
+      end
+
+      context 'when payment_ref exists' do
+        it 'finds the correct documents' do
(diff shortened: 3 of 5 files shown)


app/controllers/documents_controller.rb
```diff
@@ -19,4 +19,10 @@ def letter_file_name(doc)
     letter_template = meta.dig(:template_id)
     pay_ref.to_s + '_' + letter_template.to_s + doc.extension
   end
+
+  def params_for_index
+    payment_ref = params.fetch(:payment_ref, nil)
+
+    payment_ref ? { payment_ref: payment_ref } : {}
```
cybojenix: can be simplified to `{ payment_ref: params.fetch(:payment_ref, nil) }`

`nil` is the default for `get_all_documents`

lib/hackney/letter/all_documents_use_case.rb
```diff
@@ -5,8 +5,9 @@ def initialize(cloud_storage:)
         @cloud_storage = cloud_storage
       end
 
-      def execute
-        @cloud_storage.all_documents.each do |doc|
+      def execute(payment_ref: nil)
+        @cloud_storage.all_documents(payment_ref: payment_ref)
```
cybojenix: test hasn't been added to assert the usecase can take a payment ref

spec/lib/hackney/cloud/storage_spec.rb
```diff
@@ -12,8 +12,33 @@
...
+      let(:payment_ref) { Faker::Number.number(10) }
+
+      before do
+        Hackney::Cloud::Document.create(filename: '-', uuid: SecureRandom.uuid, extension: '-', mime_type: '-',
+                                        metadata: { payment_ref: payment_ref }.to_json)
+      end
+
+      context 'when payment_ref exists' do
+        it 'finds the correct documents' do
+          expect(storage.all_documents(payment_ref: payment_ref).count).to eq(1)
```
mrosel: just a bit flaky of a test because this gives the allusion there should only be 1 when this is on a where clause. RECOMMEND using `include` instead of eq and performing triangulation

spec/lib/hackney/cloud/storage_spec.rb
```diff
@@ -12,8 +12,33 @@
...
+        end
+      end
+
+      context 'when payment_ref does not exist' do
+        it 'returns 0 entries' do
+          expect(storage.all_documents(payment_ref: 'NON-EXISTENT-PAYMENT-REF').count).to eq(0)
```
mrosel: same ^

spec/lib/hackney/cloud/storage_spec.rb
```diff
@@ -12,8 +12,33 @@
...
+      expect(Hackney::Cloud::Document).to receive(:all)
+
+      storage.all_documents
+    end
+
+    context 'when payment_ref param is used' do
+      let(:payment_ref) { Faker::Number.number(10) }
+
+      before do
+        Hackney::Cloud::Document.create(filename: '-', uuid: SecureRandom.uuid, extension: '-', mime_type: '-',
```
mrosel: please use factories like other test cases.

lib/hackney/letter/all_documents_use_case.rb
alexdesi: Yep, this use case had missing test. Sorry no time to add it, we still need to build the front end for this.
The tests for the feature is covered well anyway, there is an integration test and a unit test.

app/controllers/documents_controller.rb
```diff
@@ -19,4 +19,10 @@ def letter_file_name(doc)
     letter_template = meta.dig(:template_id)
     pay_ref.to_s + '_' + letter_template.to_s + doc.extension
   end
+
+  def params_for_index
```
mrosel: SUGGEST: renaming this to something more relevant
```suggestion
  def payment_ref_from_params
  
```

app/controllers/documents_controller.rb
alexdesi: good point, thanks.

mrosel: this is really overcomplicating things as ```params.permit(:payment_ref).slice(:payment_ref)``` would replace this entire method

spec/lib/hackney/cloud/storage_spec.rb
alexdesi: I like it!

mrosel:
@alexdesi stop force-pushing on collaborated branches.

app/controllers/documents_controller.rb
alexdesi: Yep, I had the feeling this could be improved.
Anyway, at the end we did not need to filter out the params, since the method is handling nil.

spec/lib/hackney/cloud/storage_spec.rb
alexdesi: done, I've also used subject, that make things clearer

alexdesi:
Thanks for the feedback, should be fine now!

PR #172, opened 2019-05-02
add a task to generate failed message reports

cybojenix:
This changes the manual process that depends on PaperTrail to directly use govnotify.

The current working logic is:
- Tenancy refs are always in the SMS Message ref
- Extract it out for known patterns
- dump the full SMS reference if it doesn't match a known pattern
- output the phone number and (optionally) the URL to the tenancy on the website.

### DIFF
--- a/lib/hackney/notification/gov_notify_gateway.rb
+++ b/lib/hackney/notification/gov_notify_gateway.rb
@@ -63,8 +63,8 @@ module Hackney
       def get_messages(type: nil, status: nil)
         Enumerator.new do |enum|
           last_id = nil
-          while collection = fetch_messages(type: type, status: status, older_than: last_id).collection.presence
-            for response in collection
+          while (collection = fetch_messages(type: type, status: status, older_than: last_id).collection.presence)
+            collection.each do |response|
               last_id = response.id
               enum.yield response
             end
--- a/lib/hackney/notification/gov_notify_gateway.rb
+++ b/lib/hackney/notification/gov_notify_gateway.rb
@@ -60,6 +60,18 @@ module Hackney
         end
       end
 
+      def get_messages(type: nil, status: nil)
+        Enumerator.new do |enum|
+          last_id = nil
+          while collection = fetch_messages(type: type, status: status, older_than: last_id).collection.presence
+            for response in collection
+              last_id = response.id
+              enum.yield response
+            end
+          end
+        end
+      end
+
       private
 
       def client
@@ -85,6 +97,14 @@ module Hackney
       def pre_release_email(email)
         return email if @send_live_communications
         @test_email_address
+      end
+
+      def fetch_messages(type: nil, status: nil, older_than: nil)
+        client.get_notifications({
+          template_type: type,
+          status: status,
+          older_than: older_than
+        }.compact)
       end
     end
   end
--- a/lib/hackney/notification/gov_notify_gateway.rb
+++ b/lib/hackney/notification/gov_notify_gateway.rb
@@ -61,15 +61,13 @@ module Hackney
       end
 
       def get_messages(type: nil, status: nil)
-        Enumerator.new do |enum|
-          last_id = nil
-          while (collection = fetch_messages(type: type, status: status, older_than: last_id).collection.presence)
-            collection.each do |response|
-              last_id = response.id
-              enum.yield response
-            end
-          end
+        messages = []
+        last_id = nil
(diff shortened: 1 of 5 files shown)


lib/hackney/notification/gov_notify_gateway.rb
```diff
@@ -60,6 +60,18 @@ def get_templates(type: nil)
...
+        Enumerator.new do |enum|
+          last_id = nil
+          while (collection = fetch_messages(type: type, status: status, older_than: last_id).collection.presence)
+            collection.each do |response|
+              last_id = response.id
+              enum.yield response
+            end
+          end
+        end
+      end
```
alexdesi: This sounds quite tricky to read. I spent 10 min to understand it :)
This is an enumerator with 2 nested cycles, and then it uses `map` in the caller.
I think this could be simplified. Do we really need an enumerator?
It could return the elements already mapped .

It'd be nice to have feedback from @mrosel, @elena-vi or @cob16

lib/tasks/bau/failed_messages_report.rake
```diff
@@ -0,0 +1,25 @@
+require 'cgi'
+
+namespace :bau do
+  desc 'generate report of tenancy references that did not receive an SMS'
+  task :failed_messages_report, [:tenancies_uri] do |_t, args|
+    tenancies_uri = args.tenancies_uri || '/tenancies'
+    tenancy_ref_regex = '\\d+/\\d+'
+    uuid_regex = '[\\da-f\\-]{32,36}'
+    matches = [
+      # SendGreenInArrearsMsgJob-#{case_priority.tenancy_ref}-#{SecureRandom.uuid}
```
alexdesi: maybe we do not need this comments

lib/tasks/bau/failed_messages_report.rake
```diff
@@ -0,0 +1,25 @@
...
+      Regexp.new("^\\S+-(#{tenancy_ref_regex})-#{uuid_regex}$"),
+      # manual_#{tenancy.ref}
```
alexdesi: and this

lib/tasks/bau/failed_messages_report.rake
```diff
@@ -0,0 +1,25 @@
...
+      Regexp.new("^\\S+_(#{tenancy_ref_regex})$")
+    ]
+
+    use_cases = Hackney::Income::UseCaseFactory.new
+    use_cases.get_failed_sms_messages.execute.each do |message|
+      reference = message[:reference]
+      tenancy_refs = matches.map { |r| r.match(reference)&.values_at(1) }.flatten.compact
+      tenancy_ref = tenancy_refs.first || reference
```
alexdesi: you could clean up a bit with a private method: `def tenancy_ref (message_ref)`  (or `get_tenancy_reg`)

cybojenix: I looked into this, rake doesn't have the concept of a private method.
All methods are global, so introducing methods will result in collisions down the road

lib/tasks/bau/failed_messages_report.rake
cybojenix: I think these give context for what the references should look like. Regex isn't everyone's cup of tea, so will leave the examples in.

lib/hackney/notification/gov_notify_gateway.rb
cybojenix: For our usecase, no we don't need an enumerator.
Especially given the way it's used, the whole thing is put in memory.

I've flattened this out to lean on the side of readability

lib/hackney/notification/gov_notify_gateway.rb
```diff
@@ -60,6 +60,16 @@ def get_templates(type: nil)
         end
       end
 
+      def get_messages(type: nil, status: nil)
+        messages = []
+        last_id = nil
+        while (collection = fetch_messages(type: type, status: status, older_than: last_id).collection.presence)
```
mrosel: recommend not using a `while loop`. It's considered taboo in many ruby circles.

cybojenix: the while loop fits the need the best, as the response is paginated with no metadata.
Unless you can think of an alternate solution?

mrosel: yeah, I had missed that you don't know it ends unless there's no more left.

cybojenix:
Aye, this is very much a temporary script for during the support period. The final feature should be using delivery reports and properly tying to the action diary

PR #174, opened 2019-05-02
Template clean up

elena-vi:
A few templating and letter fixes here: 
- international address now fail validation
- `arrears_letter_1_date` date is now being fished out of our db if it's there
- removed templates which are not going to be used by the users


app/models/hackney/service_charge/letter.rb
```diff
@@ -46,9 +46,19 @@ def validate_mandatory_fields(letter_params)
...
         letter_params[:lessee_short_name] = letter_params[:lessee_full_name] unless letter_params[:lessee_short_name].present?
         letter_params
       end
+
+      def fetch_previous_letter_date(payment_ref)
+        sent_letter1 = Hackney::Cloud::Document
+                       .where("JSON_EXTRACT(metadata, '$.template.name') Like  ?", '%Letter 1%')
+                       .where("JSON_EXTRACT(metadata, '$.payment_ref') = ?", payment_ref)
+
+        sent_letter1.any? ? sent_letter1.last.updated_at : ''
```
elena-vi:
```suggestion
        sent_letter1.any? ? sent_letter1.last.updated_at.strftime('%d %B %Y') : ''
```

PR #175, opened 2019-05-03
fixing decimal places

elena-vi:
- Soraya and Guy have requested that address line 1 is required and 3 isn't 
- balance is being rounded to 2 decimal places 
- names and addresses are being stripped

### DIFF
--- a/app/models/hackney/service_charge/letter.rb
+++ b/app/models/hackney/service_charge/letter.rb
@@ -2,7 +2,7 @@ module Hackney
   module ServiceCharge
     class Letter
       MANDATORY_LETTER_FIELDS = %i[payment_ref lessee_full_name
-                                   correspondence_address2 correspondence_address3
+                                   correspondence_address1 correspondence_address2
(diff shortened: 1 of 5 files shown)


app/models/hackney/service_charge/letter.rb
```diff
@@ -2,7 +2,7 @@ module Hackney
   module ServiceCharge
     class Letter
       MANDATORY_LETTER_FIELDS = %i[payment_ref lessee_full_name
-                                   correspondence_address2 correspondence_address3
+                                   correspondence_address1 correspondence_address2
```
alexdesi: That does not work,
`correspondence_address1` is formed by `corr_preamble` ( the flat number/house Name)
that is missing most of the time in UH db.

alexdesi: `correspondence_address2` and `correspondence_address3` does not contain consisten data, we need a better analisys.

It might be that (correspondence_address2 || correspondence_address3) is mandatory

PR #184, opened 2019-05-21
Revert "fix checking state when document is accepted"

cybojenix:
This needs a bit more work, and it's blocking a deploy to production

Reverts LBHackney-IT/lbh-income-api#182


cybojenix:
wrong one

PR #186, opened 2019-05-30
Fix address box validation

booyaa:
update style so it's always within the box


cybojenix:
Comfirmed tested against gov notify validation

PR #196, opened 2019-07-17
Increase width of lessee address box

booyaa:
**What**

a) Very long Lessee full name is causing the letter to fail validation by Gov Notify. Have increased the max-width for this line. 
b) Reclaim some of the whitespace at the top of the envelope window box.

**Why**

a) Resolves validation issues
b) allows additional wrapping of up to two lines, at the cost of Lessee and Sender address blocks no longer being aligned.


cob16:
Gov notify also have limits on length as well as height. Have you checked the documentation that this new length is valid?

PR #197, opened 2019-07-17
Update copy in letter templates

booyaa:
Details of copy can be found in the Trello card. There are no changes to the lessee address box so there should be no validation concerns.


booyaa:
> I'd suggest saving this text in the layouts folder and adding it as a partial here: https://github.com/LBHackney-IT/lbh-income-api/blob/master/lib/hackney/pdf/preview_generator.rb#L8
> 
> This would make it easier to change the text only in one place in the future

That would've been a great idea, except the wording has XX days and this is dependant on if the letter is type 1 or 2. Will probably suggest that we come up with a better solution so we can extract this as a sub-template.

PR #199, opened 2019-07-25
Swap from individuals to teams

booyaa:
**What**

Moves away from named individuals to use the `Madetech` (sic) team within the Hackney org (`LBHackney-IT`).

**Why**

This will stop unnecessary code churn when people join and leave the project. 

**Concerns**

Would recommend pruning the list of users in the `Madetech` team or perhaps creating a new team called `ManageArrears`?

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -5,4 +5,4 @@
 # @global-owner1 and @global-owner2 will be requested for
 # review when someone opens a pull request.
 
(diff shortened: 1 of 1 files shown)


booyaa:
After discussing with @cob16 we've agreed to remove this restriction altogether. It made sense when there was a full delivery team working on the code.

PR #203, opened 2019-09-16
Improve README project setup

richrace:
### Why?

Getting the project to boot up the first time needed to some extra information that wasn't obvious.

### What

- Added `.env.sample` to help highlight the missing information

### Guidance to Review

- Spelling
- Is this actually useful?

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,15 +21,18 @@ We employ a variant of Clean Architecture, borrowing from [Made Tech Flavoured C
 1. Install [Docker][docker-download].
 2. Get a hackney aws account (see maintainers)
 3. Clone this repository.
-4. Login to ecr [Universal Housing Simulator][github-uh-simulator]:
+4. Login to ecr [Universal Housing Simulator][github-uh-simulator] [1]:
 ```bash
 $ aws configure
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -21,16 +21,19 @@ We employ a variant of Clean Architecture, borrowing from [Made Tech Flavoured C
...
 $ aws ecr get-login --no-include-email --region $AWS_DEFAULT_REGION | sh
 ```
-5. Run setup
+5. Duplicate `.env.sample` to `.env` and replace placeholders with valid secrets.
+6. Run setup
 ```bash
 make setup
 ```
 
+[1] Need to login to your AWS Hackney account; if you have another you will need to use Profiles
```
kjdchapman: It might be clearer to cover this as part of l24 above, going into less detail (I'm happy to leave the burden of understanding Profiles on the reader):

> 4. Login using your AWS Hackney account, so that you can pull the Universal Housing Simulator][github-uh-simulator] image from ECR.

richrace: We stumbled over this as the error you get is not exactly helpful.

PR #207, opened 2019-09-23
add nosp dates to universal housing simulator

elena-vi:
@maysakanoni @georgeschena


lib/hackney/income/tenancy_prioritiser/universal_housing_criteria.rb
```diff
@@ -88,6 +97,14 @@ def balance
           attributes.fetch(:current_balance).to_f
         end
 
+        def nosp_served_date
+          attributes.fetch(:nosp_served_date).to_date if attributes.fetch(:nosp_served_date)
```
kjdchapman: I think `fetch` will blow up if it doesn't find the key, rather than returning something falsey.

elena-vi:
resolved here https://github.com/LBHackney-IT/lbh-income-api/pull/210

PR #211, opened 2019-09-30
Add weekly balance  and latest communication action and date to sync

maysakanoni:
**What?**
-  Extracted weekly rent for tenancy from UH

- Saved this on income-collection db

- Extracted last_communication_action (code) from the araction table is UH

- Extracted last_communication_date from the araction table is UH

**Why?**

- To do weekly sync and calculate priority score

- Save last_communication_action and last_communication_date to income-api ab

### DIFF
--- a/lib/hackney/income/tenancy_prioritiser/universal_housing_criteria.rb
+++ b/lib/hackney/income/tenancy_prioritiser/universal_housing_criteria.rb
@@ -10,6 +10,8 @@ module Hackney
             DECLARE @NospActionDiaryCode VARCHAR(60) = ?
             DECLARE @PaymentTypes table(payment_type varchar(3))
             INSERT INTO @PaymentTypes VALUES ('RBA'), ('RBP'), ('RBR'), ('RCI'), ('RCO'), ('RCP'), ('RDD'), ('RDN'), ('RDP'), ('RDR'), ('RDS'), ('RDT'), ('REF'), ('RHA'), ('RHB'), ('RIT'), ('RML'), ('RPD'), ('RPO'), ('RPY'), ('RQP'), ('RRC'), ('RRP'), ('RSO'), ('RTM'), ('RUC'), ('RWA')
+            DECLARE @CommunicationTypes table(communication_types varchar(60))
+            INSERT INTO @CommunicationTypes VALUES ('C'), ('MML'), ('S0A'), ('REF'), ('ZW1'), ('ZW2'), ('ZW3'), ('MW1'), ('MW2'), ('MW3'), ('LF1'), ('LF2'), ('LL1'), ('LL2'), ('LS1'), ('LS2'), ('SMS'), ('GAT'), ('GAE'), ('GME'), ('GMS'), ('AMS')
 
             DECLARE @CurrentBalance NUMERIC(9, 2) = (SELECT cur_bal FROM [dbo].[tenagree] WITH (NOLOCK) WHERE tag_ref = @TenancyRef)
             DECLARE @LastPaymentDate SMALLDATETIME = (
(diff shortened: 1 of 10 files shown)


lib/hackney/income/tenancy_prioritiser/universal_housing_criteria.rb
```diff
@@ -97,6 +102,10 @@ def balance
           attributes.fetch(:current_balance).to_f
         end
 
+        def weekly_rent
+          attributes.fetch(:weekly_rent).to_f
```
kjdchapman: Is it OK to use floats for money here?

elena-vi: Our thinking was to go with a float here because the `current_balance` is a float. 
I'd feel quite uncomfortable handling money in one way here and in another somewhere else.

kjdchapman: > I'd feel quite uncomfortable handling money in one way here and in another somewhere else.
Same! 

Just wanted to share the thought out loud, this might cause us an issue at _unspecified point in the future_, not sure how to stop that.

lib/hackney/income/tenancy_prioritiser/universal_housing_criteria.rb
```diff
@@ -119,6 +146,16 @@ def days_since_last_payment
           day_difference(Date.today, attributes.fetch(:last_payment_date))
         end
 
+        def last_communication_action
+          return nil if attributes.fetch(:last_communication_action).nil?
```
kjdchapman: Oops, missed this had been added on.

`fetch` throws a `KeyError` if it doesn't find the key, so this nil check won't ever be hit.

I'd be happy just removing this guard clause and letting [l151](https://github.com/LBHackney-IT/lbh-income-api/pull/211/files#diff-266da5637de8cfb8fb901a365e8bd132R151) do all the work.

lib/hackney/income/tenancy_prioritiser/universal_housing_criteria.rb
```diff
@@ -119,6 +146,16 @@ def days_since_last_payment
...
+          attributes.fetch(:last_communication_action)
+        end
+
+        def last_communication_date
+          return nil if attributes.fetch(:last_communication_date).nil?
```
kjdchapman: [Same comment as above](https://github.com/LBHackney-IT/lbh-income-api/pull/211/files#r330457344).

lib/hackney/income/tenancy_prioritiser/universal_housing_criteria.rb
```diff
@@ -147,13 +147,13 @@ def days_since_last_payment
         end
 
         def last_communication_action
-          return nil if attributes.fetch(:last_communication_action).nil?
-          attributes.fetch(:last_communication_action)
+          return nil if attributes[:last_communication_action].nil?
```
kjdchapman: I think you can just skip this line - the line below will return `nil` if it's `nil` anyway?

spec/lib/hackney/income/tenancy_prioritiser/universal_housing_criteria_spec.rb
```diff
@@ -123,6 +131,52 @@
...
+      it { is_expected.to be_nil }
+    end
+
+    context 'when in communication with the tenant' do
+      before {
+        create_uh_action(tenancy_ref: tenancy_ref, code: 'S0A', date: Date.yesterday)
+        create_uh_action(tenancy_ref: tenancy_ref, code: 'MML', date: Date.today)
+      }
+
+      it 'return the latest communication code' do
```
maysakanoni:
```suggestion
      it 'return the latest communication date' do
```

PR #213, opened 2019-10-08
add logic to classify tenancies

maysakanoni:

PR #214, opened 2019-10-08
saving classifications

elena-vi:
Part two of:
https://github.com/LBHackney-IT/lbh-income-api/pull/213

once classifications have been calculated, this needs to be saved into the database

### DIFF
--- a/app/models/hackney/income/models/case_priority.rb
+++ b/app/models/hackney/income/models/case_priority.rb
@@ -2,6 +2,8 @@ module Hackney
   module Income
     module Models
       class CasePriority < ApplicationRecord
+        enum classification: %i[no_action send_letter_two send_letter_one send_first_SMS send_NOSP send_warning_letter]
+
         validates :case_id, presence: true, uniqueness: true
 
         before_validation :create_case_with_tenancy_ref
--- a/db/migrate/20191008164749_add_classification_to_case_priority.rb
+++ b/db/migrate/20191008164749_add_classification_to_case_priority.rb
@@ -1,0 +1,5 @@
+class AddClassificationToCasePriority < ActiveRecord::Migration[5.2]
+  def change
+    add_column :case_priorities, :classification, :integer
+  end
+end
--- a/lib/hackney/income/stored_tenancies_gateway.rb
+++ b/lib/hackney/income/stored_tenancies_gateway.rb
@@ -8,6 +8,8 @@ module Hackney
           criteria,
           weightings
         )
+        classification_usecase = Hackney::Income::TenancyPrioritiser::TenancyClassification.new(criteria)
+
         begin
           GatewayModel.find_or_create_by(tenancy_ref: tenancy_ref).tap do |tenancy|
             tenancy.update(
@@ -39,7 +41,8 @@ module Hackney
               nosp_expiry_date: criteria.nosp_expiry_date,
               last_communication_action: criteria.last_communication_action,
               last_communication_date: criteria.last_communication_date,
-              active_nosp: criteria.active_nosp?
+              active_nosp: criteria.active_nosp?,
+              classification: classification_usecase.execute
             )
           end
         rescue ActiveRecord::RecordNotUnique
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2019_10_01_160536) do
+ActiveRecord::Schema.define(version: 2019_10_08_164749) do
 
   create_table "case_priorities", force: :cascade do |t|
     t.string "tenancy_ref"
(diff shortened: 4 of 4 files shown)


db/migrate/20191008164749_add_classification_to_case_priority.rb
```diff
@@ -0,0 +1,5 @@
+class AddClassificationToCasePriority < ActiveRecord::Migration[5.2]
+  def change
+    add_column :case_priorities, :classification, :integer
```
maysakanoni: This is more of a question than a comment, do we save symbols as integers in the db?

elena-vi: ✨enum magic ✨
https://naturaily.com/blog/ruby-on-rails-enum

lib/hackney/income/stored_tenancies_gateway.rb
```diff
@@ -8,6 +8,7 @@ def store_tenancy(tenancy_ref:, priority_band:, priority_score:, criteria:, weig
           criteria,
           weightings
         )
+        classification_usecase = Hackney::Income::TenancyPrioritiser::TenancyClassification.new(criteria)
```
richrace: Style suggestion: Whitespace between line `11-12` 🙃

app/models/hackney/income/models/case_priority.rb
```diff
@@ -2,6 +2,8 @@ module Hackney
   module Income
     module Models
       class CasePriority < ApplicationRecord
+        enum classification: %i[no_action send_letter_two send_letter_one send_first_SMS send_NOSP send_warning_letter]
```
richrace: As these are used here https://github.com/LBHackney-IT/lbh-income-api/blob/257e89607b77765976a18a86611e7f62533a03f0/lib/hackney/income/tenancy_prioritiser/tenancy_classification.rb#L11-L16 could we look at making them either Constants or using enum values `Income::Models::CasePriority.classifications[:no_action]` ?

This should make it easier if we want to add any in the future.

lib/hackney/income/stored_tenancies_gateway.rb
```diff
@@ -8,6 +8,8 @@ def store_tenancy(tenancy_ref:, priority_band:, priority_score:, criteria:, weig
...
```
kjdchapman: I know it hasn't been introduced in this PR, but I don't "get" this use case name quickly. I'm also a little confused by the namespace. 

I would have expected this to be something like `Hackney::Income::UseCase::ClassifyTenancy`.

kjdchapman: Relying directly on a Use Case in a Gateway will tie high-level policy to low-level details, which could make this difficult to work with in the future. 

I'm worried about both the components involved (I'd expect a use case to use a gateway, not the other way around) and the lack of dependency injection.

I haven't looked at the entire architecture around this component, so I may be missing something.

lib/hackney/income/tenancy_prioritiser/tenancy_classification.rb
```diff
@@ -7,12 +7,15 @@ def initialize(criteria)
...
         def execute
-          return :send_NOSP if send_nosp?
-          return :send_warning_letter if send_warning_letter?
-          return :send_letter_two if send_letter_two?
-          return :send_letter_one if send_letter_one?
-          return :send_first_SMS if send_sms?
-          :no_action
+          case_classifications = Hackney::Income::Models::CasePriority.classifications
+
+          return case_classifications[:send_NOSP]           if send_nosp?
```
kjdchapman: As discussed irl, this strikes me as the "wrong" control flow to express this logic, but it isn't particularly unclear when read. 

I'd be happy with us releasing this, and I'm happier now it's tied back to the enum of the relevant class.

spec/lib/hackney/income/tenancy_prioritiser/tenancy_classification_spec.rb
```diff
@@ -5,6 +5,7 @@
 
   let(:criteria) { Stubs::StubCriteria.new }
   let(:assign_classification) { described_class.new(criteria) }
+  let(:classifications) { Hackney::Income::Models::CasePriority.classifications }
```
kjdchapman: I'm not sure if we should include this in the tests or not - I'd lean towards no. 

Including it here ties the tests to a particular classes' details, and unlike the production code, I'm not worried about getting the symbols wrong (as my tests would then fail).

richrace: So using  `Hackney::Income::Models::CasePriority.classifications[:no_action]` will return the enum value (`0` in this case) so the tests would read something like `eq(0)` or `eq(1)` which are not readable.

lib/hackney/income/stored_tenancies_gateway.rb
richrace: I guess this is the same problem as Line 7?

spec/lib/hackney/income/tenancy_prioritiser/tenancy_classification_spec.rb
kjdchapman: Ah, fair enough. Is it a design smell that the result of `use_case.execute` is a meaningless number?

richrace: Having a conversation with @kjdchapman we have agreed that the Use Case's return value not being readable is more of an issue that using the defined enum list.

Also, we should have a spec covering the case classification in `spec/lib/hackney/income/stored_tenancies_gateway_spec.rb`

lib/hackney/income/stored_tenancies_gateway.rb
richrace: Discussed with @kjdchapman and we're going to leave this alone for now and it's probably quite a bit to tackle which is not the scope of this PR

PR #215, opened 2019-10-11
LBA template

Miles-Alford:
WHAT: add template for Letter Before Action to API
WHY: so the front-end is able to generate a LBA template

### DIFF
--- a/lib/hackney/pdf/templates/letter_before_action.erb
+++ b/lib/hackney/pdf/templates/letter_before_action.erb
@@ -1,0 +1,122 @@
+<div class="top_space"></div>
+
+<div class="right">
+<strong>Leasehold & Right To Buy Service</strong> <br>
+  Hackney Service Centre <br>
+  1 Hillman Street <br>
+  London E8 1DY
+
+<p>
+  020 8356 2299 <br>
+  service.charges@hackney.gov.uk <br>
+  <strong>Service Charge Account Number: <%= @letter.payment_ref %> <br></strong>
+  Our Ref: LBA/ST
+  <%= @sending_date + 1.day %>
+</p>
+</div>
+
+<div class="lessee_address">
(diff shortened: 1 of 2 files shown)


kjdchapman:
What's the easiest way for me to see an example pdf this generates?

lib/hackney/pdf/templates/letter_before_action.erb
```diff
@@ -0,0 +1,247 @@
...
+  Hackney Service Centre <br>
+  1 Hillman Street <br>
+  London E8 1DY
+
+<p>
+  020 8356 2299 <br>
+  service.charges@hackney.gov.uk <br>
+  <strong>Service Charge Account Number: <%= @letter.payment_ref %> <br></strong>
+  Our Ref: LBA/ST
+  <%= @sending_date + 1.day %>
```
elena-vi: `@sending_date` is already today's date + 1
also `@sending_date` is a string and adding time to this will cause this error
```
TypeError: no implicit conversion of ActiveSupport::Duration into String

```

lib/hackney/pdf/templates/letter_before_action.erb
```diff
@@ -0,0 +1,247 @@
...
+
+<h2>What if I am currently experiencing financial difficulties?</h2>
+<p>
+We are currently aware of a valid reason for your non_payment. If you are facing financial
+difficulties, please complete the enclosed forms to allow us to assess your current 
+circumstances and the options available to uo
+<p>
+
+<h2>If you are intending to return these forms, please ensure they are received by the 
+Council no later than <%= @sending_date + 31.days %> </h2>
```
elena-vi: I think it's inappropriate to do the date calculation here. 
It would be much more appropriate to declare a new date variable [here](https://github.com/LBHackney-IT/lbh-income-api/blob/master/lib/hackney/pdf/preview_generator.rb) and interpolate it into the template.

PR #216, opened 2019-10-14
Sync Patch Code

richrace:
### What

- Sync Patch Code from UH 

### Why

- To allow us to show which case is in which patch

### DIFF
--- a/spec/lib/hackney/income/tenancy_prioritiser/universal_housing_criteria_spec.rb
+++ b/spec/lib/hackney/income/tenancy_prioritiser/universal_housing_criteria_spec.rb
@@ -431,33 +431,31 @@ describe Hackney::Income::TenancyPrioritiser::UniversalHousingCriteria, universa
   end
 
   describe '#patch_code' do
-    context 'when there is a patch code' do
-      let(:patch_tenancy_code) { '100000/11' }
-      let(:patch_code) { 'E01' }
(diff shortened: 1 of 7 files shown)


spec/lib/hackney/income/tenancy_prioritiser/universal_housing_criteria_spec.rb
```diff
@@ -429,4 +429,38 @@
...
+      end
+    end
+
+    context 'when there is no property' do
+      it 'is nil' do
+        expect(criteria.patch_code).to be_nil
+      end
+    end
+
+    context 'when there is a property but no patch code' do
```
kjdchapman: The before/context blocks are a little confusing for me. I'd expect this to be structured something like:

```ruby
context 'with an existing property reference' do
  context 'with a patch code' do; end
  context 'without a patch code' do; end
end

context "with a property reference that doesn't resolve"
end

richrace: Agreed.

richrace: Agreed.

kjdchapman: Nice one, 1eedce7529 addresses this well :tada:

PR #218, opened 2019-10-15
Add case classification to my cases

richrace:
### What 

- Add classification to my cases for next recommended action

### Why

- So the front end can tell the user what to do next

### DIFF
--- a/spec/lib/hackney/income/view_my_cases_spec.rb
+++ b/spec/lib/hackney/income/view_my_cases_spec.rb
@@ -125,7 +125,9 @@ describe Hackney::Income::ViewMyCases do
                                            active_agreement: tenancy_priority_factors.fetch(:active_agreement),
                                            broken_court_order: tenancy_priority_factors.fetch(:broken_court_order),
                                            nosp_served: tenancy_priority_factors.fetch(:nosp_served),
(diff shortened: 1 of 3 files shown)


spec/lib/hackney/income/view_my_cases_spec.rb
```diff
@@ -125,7 +125,9 @@
                                            active_agreement: tenancy_priority_factors.fetch(:active_agreement),
                                            broken_court_order: tenancy_priority_factors.fetch(:broken_court_order),
                                            nosp_served: tenancy_priority_factors.fetch(:nosp_served),
-                                           active_nosp: tenancy_priority_factors.fetch(:active_nosp)
+                                           active_nosp: tenancy_priority_factors.fetch(:active_nosp),
+
+                                           classification: tenancy_priority_factors.fetch(:classification)
```
kjdchapman: I know this covers more than just your feature, but is there any triangulation in these test suites? Looks like at the moment this would all pass with hard-coded values.

richrace: I think this test is only testing that things are wired up correctly and the actual data of any "attribute" is irrelevant.

PR #223, opened 2019-10-17
Generate lba

maysakanoni:


app/models/hackney/service_charge/letter.rb
```diff
@@ -96,6 +102,10 @@ def fetch_previous_letter_date(payment_ref)
 
         sent_letter1.any? ? sent_letter1.last.updated_at.strftime('%d %B %Y') : ''
       end
+
+      def calculate_lba_balance(arrears_balance, money_judgement, charging_order, bal_dispute)
```
elena-vi: we need to think about what would happen if this method returned a negative number (not sure if that would actually happen but I think we should cover that scenario just incase)

for example 
```ruby 
lba_balance = calculate_lba_balance(arrears_balance: 100, money_judgement: 20, charging_order: 25, bal_dispute: 25)
=> -20
```

maysakanoni: The money judgement and charging order are already accounted for in the arrears balance which is why we are subtracting it.

app/models/hackney/service_charge/letter.rb
```diff
@@ -31,12 +31,16 @@ def initialize(params)
         @lba_expiry_date = validated_params[:lba_expiry_date]
         @original_lease_date = validated_params[:original_lease_date]
         @date_of_current_purchase_assignment = validated_params[:date_of_current_purchase_assignment]
-        @original_leaseholders = validated_params[:original_leaseholders]
+        @original_leaseholders = 'the original leaseholder'
```
richrace: Is this meant to be here?

.circleci/config.yml
```diff
@@ -105,12 +105,7 @@ workflows:
   continuous_delivery:
     jobs:
       - check
-      - staging_release:
-          requires:
-            - check
-          filters:
-            branches:
-              only: master
```
richrace: This needs to be put back in before Merging

Miles-Alford: Will do, just going to check if the changes work correctly and will be put back afterwards

app/models/hackney/service_charge/letter.rb
```diff
@@ -96,6 +105,10 @@ def fetch_previous_letter_date(payment_ref)
 
         sent_letter1.any? ? sent_letter1.last.updated_at.strftime('%d %B %Y') : ''
       end
+
+      def calculate_lba_balance(arrears_balance, money_judgement)
+        arrears_balance.to_f - money_judgement.to_f
```
richrace: I think we should be using `BigDecimal` over `float` when dealing with any money calculations.

https://blog.bigbinary.com/2013/01/14/handling-money-in-ruby.html

spec/models/hackney/service_charge/letter_spec.rb
```diff
@@ -55,4 +60,43 @@
...
+      expected_total_collectable_arrears_balance = format('%.2f', balance - money_judgement)
+      expect(letter.lba_balance).to eq(expected_total_collectable_arrears_balance)
+    end
+  end
+
+  describe 'tenure type' do
+    let(:letter) { described_class.new(letter_params) }
+
+    context 'with a tenure type of FRS' do
+      let(:tenure_type) { 'FRS' }
```
richrace: Would it be possible to make `FRS`, `LEA` and `SHO` (+ any others) into constants?

PR #225, opened 2019-10-21
Add filter for cases with unassigned Patch Codes

richrace:
### Why

- As a user I need to filter the Worktray to the Patch I'm currently working on

### What

- Add special filter for cases without a Patch

### Guidance for view

- How do we feel about the magic `'unassigned'` string that has to match the front end?

/cc @kjdchapman @bjpirt


richrace:
So the options:

1. All - do not filter on a patch
2. A given patch - filter on the given patch
3. Unassigned - filter for cases without a patch explicitly

PR #226, opened 2019-10-21
Do not filter worktray by User

richrace:
### Why

- As a User, I want to see all cases in a Patch.
- As a User, I am not assigned cases on Priority but by Patch

### What

- Remove filtering by User on the Worktray

### Notes
- Todo: Remove assigning Cases to Users in the Sync


richrace:
If we're going to be adding roles/permissions is it wise to actually do all of the PR?

Should I change this PR to just remove these lines: https://github.com/LBHackney-IT/lbh-income-api/pull/226/files#diff-de386e1a26981648585f370d208bca3cL69-L72


/cc @kjdchapman @bjpirt

PR #227, opened 2019-10-22
Fetch court_outcome from uh

alex-demetriou:
### Why
* As an Income Collection officer I need to record and view what the outcome of the court hearing was.

### What
* Added court_outcome to the payload


lib/hackney/income/tenancy_prioritiser/universal_housing_criteria.rb
```diff
@@ -154,6 +158,10 @@ def courtdate
           attributes[:courtdate].to_date
         end
 
+        def court_outcome
+          attributes[:court_outcome]
```
georgeschena: Can the `court_outcome` be nil?

PR #228, opened 2019-10-23
Update codeowners - add Richard, remove Lia

georgeschena:


CODEOWNERS
```diff
@@ -1 +1 @@
-* @elena-vi @kjdchapman @georgeschena @maysakanoni @LBHLTarar @Miles-Alford
+* @elena-vi @kjdchapman @georgeschena @maysakanoni @LBHLTarar @Miles-Alford @richrace
```
Miles-Alford: Shouldn't we remove Lia from the list?

PR #229, opened 2019-10-23
Add Request specs

richrace:
### Why

- As a developer I need to have confidence my changes do not break things

### What 

- Add end-to-end specs for `LettersController#create` action
- Add end-to-end specs for `LettersController#send_letter` action

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -32,6 +32,7 @@ ActiveRecord::Migration.maintain_test_schema!
 
 RSpec.configure do |config|
   config.include UniversalHousingHelper, type: :controller
+  config.include UniversalHousingHelper, type: :request
   config.include UniversalHousingHelper, universal: true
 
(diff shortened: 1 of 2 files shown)


kjdchapman:
Just to re-iterate our conversation in-person, these are great - and they've exposed some slightly surprisingly behaviour/design choices in this API, which is also great.

Are you looking for any further specific feedback to help this move from draft to something mergeable?

richrace:
So we agree that these tests add value. I'm happy for this to be merged in.

But:

> Are you looking for any further specific feedback to help this move from draft to something mergeable?

Are we happy with this POC and happy with the example they are setting? E.g. do these tests go into too much detail that if I was writing a new feature, I would not want to go this deep.

kjdchapman:
We talked in-person - I believe your concern was about these being more "characterization"-style tests than "normal" request specs.

I'm still happy for them to get merged in - I think it'd be good to talk us all through what you've done here, too.

PR #231, opened 2019-10-28
Add missing values for hash

georgeschena:
Quick fix for failing tests due to missing values


richrace:
Changes on master that branch wasn't aware of.

PR #233, opened 2019-10-29
Extract pdf generation and upload

georgeschena:
### WHAT
Refactoring letters controller and underlying methods

### WHY
To allow us to decouple existing code to allow us to download LBA's more easily


app/controllers/letters_controller.rb
```diff
@@ -6,18 +6,43 @@ def get_templates
   end
 
   def create
-    render json: pdf_use_case_factory.get_preview.execute(
+    letter_data = UseCases::ViewLetter.new.execute
+    _uuid = UseCases::SaveToCache.new(cache: Rails.cache).execute(data: letter_data)
```
richrace: This is currently not doing anything.

We should remove this code in this PR as we will add it in shortyl

lib/use_cases/view_letter.rb
```diff
@@ -0,0 +1,5 @@
+module UseCases
+  class ViewLetter
```
richrace: Can delete this file in this PR

spec/lib/hackney/income/sql_legal_cases_gateway_spec.rb
```diff
@@ -50,7 +50,7 @@
       end
     end
 
-    it 'returns the tenancy refs for that patch' do
+    xit 'returns the tenancy refs for that patch' do
```
richrace: Should unskip this test

spec/requests/downloading_a_letter_pdf_spec.rb
```diff
@@ -0,0 +1,22 @@
+require 'rails_helper'
+
+xdescribe 'downloading a letter PDF' do
```
richrace: Why is this test skipping?

PR #236, opened 2019-11-01
Do not store User ID in metadata for Document

richrace:
### Why

- So we can remove any reference to the User Model in a later change.

### What 

- Denormalise the User's Name and Email onto the Document Model. 

### Blockers

- Should wait for the large refactor (https://github.com/LBHackney-IT/lbh-income-api/pull/233) to be finished before merging this in.
- This change needs to be merged before https://github.com/LBHackney-IT/lbh-income-api/pull/235

### DIFF
--- a/db/migrate/20191101093146_add_columns_to_documents.rb
+++ b/db/migrate/20191101093146_add_columns_to_documents.rb
@@ -1,0 +1,18 @@
+class AddColumnsToDocuments < ActiveRecord::Migration[5.2]
+  def change
+    add_column :documents, :username, :string
+    add_column :documents, :email, :string
+
+    Hackney::Cloud::Document.find_each do |document|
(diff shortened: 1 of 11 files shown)


db/migrate/20191101093146_add_columns_to_documents.rb
```diff
@@ -0,0 +1,18 @@
+class AddColumnsToDocuments < ActiveRecord::Migration[5.2]
+  def change
+    add_column :documents, :username, :string
+    add_column :documents, :email, :string
+
+    Hackney::Cloud::Document.find_each do |document|
+      next if document.metedata.blank?
```
bjpirt: Typo on metadata

richrace:
Need to update: `lib/hackney/letter/all_documents_use_case.rb`

PR #237, opened 2019-11-05
Remove caching functionality

georgeschena:
### WHAT
Removed the caching functionality as now not needed
### WHY
We don't have the need to cache the letter now as we use the generated pdf to preview the letter


lib/use_cases/generate_and_store_letter.rb
```diff
@@ -4,6 +4,7 @@ module UseCases
   class GenerateAndStoreLetter
     def execute(payment_ref:, template_id:, user_id:)
       pdf_use_case_factory = Hackney::PDF::UseCaseFactory.new
+      letter_use_case_factory = Hackney::Letter::UseCaseFactory.new
 
       json = pdf_use_case_factory.get_preview.execute(
```
richrace: Probably want to make `json` a bit more descriptive.

Something like `letter_data`

PR #240, opened 2019-11-06
Use google auth and remove user

richrace:
### Why

- We are not the authority for Users
- Do not store a 'cache' of users

### What

- Remove references to `User`
- Let the front end send `Username` and other fields rather than us looking up User's details.


app/controllers/my_cases_controller.rb
```diff
@@ -1,9 +1,8 @@
 class MyCasesController < ApplicationController
```
bjpirt: Should this be refactored just to be CasesController now? Since there is no longer a concept of _My_ cases

lib/hackney/income/view_my_cases.rb
```diff
@@ -8,32 +8,30 @@ def initialize(tenancy_api_gateway:, stored_tenancies_gateway:)
         @stored_tenancies_gateway = stored_tenancies_gateway
       end
 
-      def execute(user_id:, page_number:, number_per_page:, filters: {})
-        number_of_pages_for_user = @stored_tenancies_gateway.number_of_pages_for_user(
-          user_id: user_id,
+      def execute(page_number:, number_per_page:, filters: {})
```
bjpirt: Like the previous comment - should this become ViewCases ?

app/controllers/my_cases_controller.rb
richrace: Yea - wondered if the change should be done outside of this PR as renaming the file etc would make it harder to see the actual changes.

bjpirt: Fair point

app/controllers/cases_controller.rb
```diff
@@ -1,39 +1,35 @@
...
+  def cases_params
     params.require(REQUIRED_INDEX_PARAMS)
     allowed_params = params.permit(REQUIRED_INDEX_PARAMS + %i[is_paused patch recommended_actions full_patch upcoming_evictions upcoming_court_dates])
 
-    allowed_params[:user_id] = allowed_params[:user_id].to_i
-
     allowed_params[:is_paused] = ActiveModel::Type::Boolean.new.cast(allowed_params[:is_paused])
     allowed_params[:full_patch] = ActiveModel::Type::Boolean.new.cast(allowed_params[:full_patch])
     allowed_params[:upcoming_court_dates] = ActiveModel::Type::Boolean.new.cast(allowed_params[:upcoming_court_dates])
+    allowed_params[:upcoming_evictions] = ActiveModel::Type::Boolean.new.cast(allowed_params[:upcoming_evictions])
```
kjdchapman: Is this related to this PR?

app/models/hackney/income/models/user.rb
```diff
@@ -2,17 +2,6 @@ module Hackney
   module Income
```
kjdchapman: Does this class need to exist? I ran tests after deleting it and they passed :man_shrugging:

lib/hackney/letter/all_documents_use_case.rb
```diff
@@ -8,20 +8,12 @@ def initialize(cloud_storage:)
...
                       .each do |doc|
-          next unless doc.metadata
-          metadata = JSON.parse(doc.metadata).deep_symbolize_keys
-          if metadata[:user_id]
-            metadata[:user_name] = user_name(metadata[:user_id])
-            doc.metadata = metadata.to_json
-          end
-        end
-      end
+          metadata = JSON.parse(doc.metadata).deep_symbolize_keys if doc.metadata
```
kjdchapman: I found this method a little harder to parse - I'm aware it's pre-existing though.

If I were to change it now, I'd write:
```
original = JSON.parse(doc.metadata)
          
modified = original.present? ? original.deep_symbolize_keys : {}
modified[:username] = doc.username
          
doc.metadata = modified.to_json
```

kjdchapman: Interestingly, just replacing this with only `@cloud_storage.all_documents(payment_ref: payment_ref)` passes all the tests - is any consumer even using the metadata?

lib/use_cases/create_document_model.rb
```diff
@@ -9,13 +9,17 @@ def initialize(document_model)
     def execute(letter_html:, uuid:, filename:, metadata:)
       extension = File.extname(filename)
 
+      metadata ||= {}
```
kjdchapman: Is this better here, or as a default parameter?

kjdchapman: I ran the tests to see if it still works with a default parameter, but they pass if you just remove this line, so :sweat_smile:

spec/lib/hackney/income/process_letter_spec.rb
```diff
@@ -1,28 +1,32 @@
 require 'rails_helper'
 
 describe Hackney::Income::ProcessLetter do
-  subject { described_class.new(cloud_storage: cloud_storage) }
+  subject { described_class.new(cloud_storage: cloud_storage_spy) }
```
kjdchapman: :wolf:  :+1:

richrace:
> With us changing how cases are viewing, do we want to soon remove references to "priorities" and "weights" (not as part of this PR)?

That is the plan to do that as a separate PR.

app/controllers/cases_controller.rb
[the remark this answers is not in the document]
richrace: Tidy up whilst I was here (https://github.com/LBHackney-IT/lbh-income-api/pull/240/files#diff-292f3032f97cf1ab08cb04b0fd402659L35)

app/controllers/letters_controller.rb
```diff
@@ -10,7 +10,8 @@ def create
     json = generate_and_store_use_case.execute(
       payment_ref: parms_for_generate_and_store[:payment_ref],
       template_id: parms_for_generate_and_store[:template_id],
-      user_id: parms_for_generate_and_store[:user_id]
+      username: parms_for_generate_and_store[:username],
```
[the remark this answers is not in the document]
richrace: 🤦‍♂

app/models/hackney/income/models/user.rb
richrace: Required for the later migration to run.

db/migrate/20191101093146_add_columns_to_documents.rb
```diff
@@ -0,0 +1,18 @@
+class AddColumnsToDocuments < ActiveRecord::Migration[5.2]
```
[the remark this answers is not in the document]
richrace: The `AddXXXToTable` is Rails magic syntax https://edgeguides.rubyonrails.org/active_record_migrations.html#creating-a-migration

Could change the `Columns` to be more reflective.

lib/hackney/cloud/storage.rb
```diff
@@ -12,14 +12,15 @@ def initialize(storage_adapter, document_model)
       def save(letter_html:, uuid:, filename:, metadata:)
         extension = File.extname(filename)
 
-        new_doc   = document_model.find_by(uuid: uuid)
-        new_doc ||= document_model.create(
+        new_doc = document_model.create(
```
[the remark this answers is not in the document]
richrace: The `new_doc   = document_model.find_by(uuid: uuid)` line was added via the refactor and slipped through. 

There is a unique index constraint on the DB column `uuid` so this should rightly blow up.

lib/hackney/letter/all_documents_use_case.rb
richrace: `original = JSON.parse(doc.metadata)` will blow up if `doc.metadata` is`nil`. Ideally `metadata` should be an ActiveRecord Serialised column and not something we manually convert so we wouldn't have to worry about parsing it.

lib/use_cases/create_document_model.rb
richrace: 1. If used as default param value, it is no longer an named param.
2. `JSON.parse` will blow up if it's `nil`

db/migrate/20191101093146_add_columns_to_documents.rb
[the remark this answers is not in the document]
kjdchapman: We talked about this in-person, clarified my concern is with the "surprise" factor of this including "backfilling" as well as "adding columns". Sounded like @richrace is happy to update the class name to try and reflect that.

app/models/hackney/income/models/user.rb
kjdchapman: Talked about this in-person, @richrace would tend to delete this later (doesn't like defining class names in the migration), I'm quite happy with that so long as it goes eventually!

lib/use_cases/create_document_model.rb
kjdchapman: Talked about this in-person - this clause might not even be needed any more. I'd prefer us to fix it (since it's an addition in this PR), even if that's just getting it under test.

PR #243, opened 2019-11-07
filter eviction date

maysakanoni:
**What?**

- Add filter for eviction date tab

**Why?**

- So caseworkers are able to see any upcoming evictions

### DIFF
--- a/lib/hackney/income/stored_tenancies_gateway.rb
+++ b/lib/hackney/income/stored_tenancies_gateway.rb
@@ -45,7 +45,8 @@ module Hackney
               classification: classification_usecase.execute,
               patch_code: criteria.patch_code,
               courtdate: criteria.courtdate,
-              court_outcome: criteria.court_outcome
+              court_outcome: criteria.court_outcome,
+              eviction_date: criteria.eviction_date
             )
           end
         rescue ActiveRecord::RecordNotUnique
@@ -137,7 +138,8 @@ module Hackney
           patch_code: model.patch_code,
           classification: model.classification,
           courtdate: model.courtdate,
-          court_outcome: model.court_outcome
+          court_outcome: model.court_outcome,
+          eviction_date: model.eviction_date
         }
       end
     end
--- a/lib/hackney/income/stored_tenancies_gateway.rb
+++ b/lib/hackney/income/stored_tenancies_gateway.rb
@@ -60,9 +60,10 @@ module Hackney
 
         query = query.offset((page_number - 1) * number_per_page).limit(number_per_page) if page_number.present? && number_per_page.present?
 
-        return query.order('eviction_date').map(&method(:build_tenancy_list_item)) if filters[:upcoming_evictions].present?
+        order_options   = 'eviction_date' if filters[:upcoming_evictions].present?
+        order_options ||= by_balance
 
-        query.order(by_balance).map(&method(:build_tenancy_list_item))
+        query.order(order_options).map(&method(:build_tenancy_list_item))
       end
 
       def number_of_pages_for_user(user_id:, number_per_page:, filters: {})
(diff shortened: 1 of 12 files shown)


lib/hackney/income/stored_tenancies_gateway.rb
```diff
@@ -82,6 +82,8 @@ def tenancies_filtered_for(user_id, filters)
           end
         end
 
+        query = query.where('eviction_date >= ?', Date.today).order('eviction_date') if filters[:upcoming_evictions].present?
```
richrace: - `Date.today` is not Time zone aware. 
- From the `schema.rb`: `t.datetime "eviction_date"` so we should query a Date Time object


```suggestion
        query = query.where('eviction_date >= ?', Time.zone.now.beginning_of_day).order('eviction_date') if filters[:upcoming_evictions].present?
```

maysakanoni:
> I think we need to ensure that we return all next actions:
> 
> See https://github.com/LBHackney-IT/lbh-income-api/pull/243/files#diff-de386e1a26981648585f370d208bca3cR104

This is now done here f01c5dc16533a5f91d81b6aa79b31cdb553d3ddb

lib/hackney/income/stored_tenancies_gateway.rb
```diff
@@ -60,6 +60,8 @@ def get_tenancies_for_user(user_id:, page_number: nil, number_per_page: nil, fil
 
         query = query.offset((page_number - 1) * number_per_page).limit(number_per_page) if page_number.present? && number_per_page.present?
 
+        return query.order('eviction_date').map(&method(:build_tenancy_list_item)) if filters[:upcoming_evictions].present?
```
richrace: `.map(&method(:build_tenancy_list_item))` is duplicated.

Could be rewritten to be:

```ruby
order_options   = 'eviction_date' if filters[:upcoming_evictions].present?
order_options ||= by_balance

query.order(order_options).map(&method(:build_tenancy_list_item))
```

maysakanoni: Nice! Ive done it here ace324b68a44d7c72851d6072df43eb66f20224b

spec/lib/hackney/income/stored_tenancies_gateway_spec.rb
```diff
@@ -436,6 +436,51 @@
...
+      last_eviction_date_created = Date.tomorrow + cases_with_upcoming_evictions - 1
+      create(:case_priority, assigned_user_id: user.id, balance: 40, eviction_date: Date.today)
+      expect(subject.first[:eviction_date]).to eq(Date.today)
+      expect(subject.last[:eviction_date]).to eq(last_eviction_date_created)
+    end
+
+    it 'can return cases in the future' do
+      create(:case_priority, assigned_user_id: user.id, balance: 40, eviction_date: Date.yesterday)
+      subject.each do |value|
+        expect(value[:eviction_date]).not_to be(Date.today)
```
richrace: This test isn't very useful. Can get rid of the `#each { ... }` and do:

`expect(subject.map { |v| v[:eviction_date] }.min).to be >= Time.zone.today`

This will ensure that the youngest date will still be in the future.

maysakanoni:
> Missing filters: https://github.com/LBHackney-IT/lbh-income-api/pull/241/files#diff-ce374463cb029b0de5da6ba98d5510cb

resolved here 197fbd43f3d3e96510591f92eee72b183147733d

spec/lib/hackney/income/stored_tenancies_gateway_spec.rb
richrace: We created a case with a date of yesterday and then check that there is not a case with date of today in the list. So this will always pass.

PR #246, opened 2019-11-11
Use Web Archive to download FreeTDS

richrace:
freetds.org website and ftp server went down. This is a workaround to 
ensure we can still develop and run our CI/CD pipeline.


richrace:
> If this fails, is it obvious what's gone wrong?
> 
> I think there's no good way to make it more obvious.

Here is a job that's failed to download FreeTDS https://circleci.com/gh/LBHackney-IT/lbh-income-api/1554

PR #248, opened 2019-11-13
Separate letter templates

elena-vi:
what letter templates are show is now determined by what group(s) the user is in


app/controllers/letters_controller.rb
```diff
@@ -3,14 +3,17 @@
 
 class LettersController < ApplicationController
   def get_templates
-    render json: pdf_use_case_factory.get_templates.execute
+    render json: pdf_use_case_factory.get_templates.execute(
+      user_groups: params_for_templates[:user_groups].split(/,/)
```
richrace: I think it might be useful to think about implementing a function in the `ApplicationController` that we can use to consistently parse the the `user_groups` in consistent way. 

Likewise, we could use the Domain model in the front end to format the groups for API transfer?

lib/hackney/pdf/get_templates.rb
```diff
@@ -1,12 +1,15 @@
 module Hackney
   module PDF
     class GetTemplates
-      def initialize(template_directory_path:)
-        @template_directory = template_directory_path
-      end
+      LEASEHOLD_SERVICES_TEMPLATE_DIRECTORY_PATH = 'lib/hackney/pdf/templates/leasehold/'.freeze
+      INCOME_COLLECTION_TEMPLATE_DIRECTORY_PATH = 'lib/hackney/pdf/templates/income/'.freeze
+      LEASEHOLD_SERVICES_GROUP = 'leasehold-services-group-1'.freeze
+      INCOME_COLLECTION_GROUP = 'income-collection-group-1'.freeze
```
richrace: These are not the real groups.

app/controllers/letters_controller.rb
```diff
@@ -3,13 +3,16 @@
...
+    )
   end
 
   def create
     json = generate_and_store_use_case.execute(
       payment_ref: params_for_generate_and_store[:payment_ref],
       template_id: params_for_generate_and_store[:template_id],
+      user_groups: params_for_generate_and_store[:user_groups],
```
richrace: The user groups here is not `split`. And going down the chain to where `get_templates` is called, it also doesn't do the `split`

lib/hackney/pdf/get_templates.rb
```diff
@@ -28,6 +31,13 @@ def get_template_name(file_path)
       def get_template_id(file_path)
         File.basename(file_path, '.*')
       end
+
+      def get_template_directory_path(groups)
```
richrace: I like this 😄

app/controllers/letters_controller.rb
elena-vi: nice catch

app/controllers/letters_controller.rb
```diff
@@ -27,7 +27,27 @@ def send_letter
...
+      user: [:id, :name, :email, groups: []]
+    )
+  end
+
+  def params_for_templates
+    params.permit(user: [:id, :name, :email, groups: []])
+  end
+
+  def user
+    # byebug
```
richrace: Debugging.

app/controllers/letters_controller.rb
```diff
@@ -27,7 +27,27 @@ def send_letter
...
+    user_params = JSON.parse(params[:user])
```
richrace: Why is this not using Strong Params?

lib/hackney/pdf/get_templates.rb
```diff
@@ -1,12 +1,13 @@
 module Hackney
   module PDF
     class GetTemplates
```
richrace: I feel that this is now a different use case e.g.

```suggestion
     class GetTemplatesForUser
```

spec/use_cases/generate_and_store_letter_spec.rb
```diff
@@ -9,14 +9,19 @@
 
   let(:use_case) { described_class.new }
   let(:use_case_output) { use_case.execute(params) }
-  let(:username) { Faker::Name.name }
+  let(:user_name) { Faker::Name.name }
```
richrace: Everywhere else is `username` apart from here.

lib/hackney/domain/user.rb
```diff
@@ -0,0 +1,15 @@
+module Hackney
```
kjdchapman: This is the biggest indicator, I think - that we're replicating an entire class across two apps.

elena-vi:
this needs the following frontend changes to work: https://github.com/LBHackney-IT/LBH-IncomeCollection/pull/243

richrace:
> I don't like spreading the policy of "who is allowed to do what?" across an app boundary. I think (as we're seeing with letters in general) it's going to make things hard to improve/maintain,
> 
> Does the team have a shared understanding/vision for what the different apps "should" do?
> 
> Are there other ways to achieve the functionality we want?

After talking in person yesterday evening, did we come to a resolution over this?

kjdchapman:
My understanding from that in-person conversation was:
- all three of us (@elena-vi , @richrace and I) saw value in moving some things from `income-api` to this repo
- we all had a concern about the amount of work it could take to, e.g., move "letters" in this way
- we wanted to talk to @bjpirt about it as well

Did I miss anything major?

I am not sure how this should impact this particular PR. We could firm up our thoughts and then have a conversation with some stakeholders about the technical approach?

bjpirt:
Let's put this in as a tech debt item to the backlog for now but let's have a discussion about what we think the right thing to do is

richrace:
Okay. So are we "okaying" this PR or does the the way the user is passed around need to change?

I'm voting for "Okay" as we need to authorise "things" on the Front end and Back end and this is a lightweight solution that can be flexible on both ends. It also will let the Front end worry about Front end things and the Back end worry about Back end things.

bjpirt:
I also vote for "Okay" and to have an architectural discussion about how we want to manage user data between the apps

PR #250, opened 2019-11-13
Commence a sync when adding a new action to a case

georgeschena:
### WHAT
Added logic to commence a sync when a new action is added to a case

### WHY
So we can have the most up to date data with regards to cases, instead of having to wait for the nightly sync

### DIFF
--- a/app/controllers/action_diary_controller.rb
+++ b/app/controllers/action_diary_controller.rb
@@ -13,6 +13,11 @@ class ActionDiaryController < ApplicationController
       render(json: { status: 'error', code: 422, message: e.message }, status: :unprocessable_entity)
       return
     end
+
+    income_use_case_factory.sync_case_priority.execute(
+      tenancy_ref: action_diary_params.fetch(:tenancy_ref)
+    )
+
     head(:no_content)
   end
 
--- a/app/controllers/action_diary_controller.rb
+++ b/app/controllers/action_diary_controller.rb
@@ -3,7 +3,7 @@ class ActionDiaryController < ApplicationController
 
   def create
     begin
-      income_use_case_factory.add_action_diary.execute(
+      income_use_case_factory.add_action_diary_and_sync_case.execute(
         tenancy_ref: action_diary_params.fetch(:tenancy_ref),
         action_code: action_diary_params.fetch(:action_code),
         comment: action_diary_params.fetch(:comment),
@@ -13,10 +13,6 @@ class ActionDiaryController < ApplicationController
       render(json: { status: 'error', code: 422, message: e.message }, status: :unprocessable_entity)
       return
     end
-
-    income_use_case_factory.sync_case_priority.execute(
-      tenancy_ref: action_diary_params.fetch(:tenancy_ref)
-    )
 
     head(:no_content)
   end
--- a/lib/use_cases/add_action_diary_and_sync_case.rb
+++ b/lib/use_cases/add_action_diary_and_sync_case.rb
@@ -1,7 +1,6 @@
 module UseCases
   class AddActionDiaryAndSyncCase
-    def initialize(prioritisation_gateway:, sync_case_priority:, action_diary_gateway:)
-      @prioritisation_gateway = prioritisation_gateway
+    def initialize(sync_case_priority:, action_diary_gateway:)
       @sync_case_priority = sync_case_priority
       @action_diary_gateway = action_diary_gateway
     end
--- a/lib/use_cases/add_action_diary_and_sync_case.rb
+++ b/lib/use_cases/add_action_diary_and_sync_case.rb
@@ -1,4 +1,3 @@
-require 'active_support'
 module UseCases
   class AddActionDiaryAndSyncCase
     def initialize(prioritisation_gateway:, sync_case_priority:, action_diary_gateway:)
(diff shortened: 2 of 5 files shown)


app/controllers/action_diary_controller.rb
```diff
@@ -13,6 +13,11 @@ def create
       render(json: { status: 'error', code: 422, message: e.message }, status: :unprocessable_entity)
       return
     end
+
+    income_use_case_factory.sync_case_priority.execute(
+      tenancy_ref: action_diary_params.fetch(:tenancy_ref)
+    )
+
```
richrace: I'm wondering if this should actually be moved into a use case? E.g. `add_action_diary_and_sync_case`?

spec/controllers/action_diary_controller_spec.rb
```diff
@@ -6,23 +6,29 @@
       username: Faker::Name.name,
       tenancy_ref: Faker::Lorem.characters(8),
       action_code: Faker::Internet.slug,
-      comment: Faker::Lorem.paragraph
+      comment: 'George was here'
```
richrace: Should this be here? 😂

spec/controllers/action_diary_controller_spec.rb
```diff
@@ -6,23 +6,29 @@
...
 
   let(:use_case_double) { double(Hackney::Tenancy::AddActionDiaryEntry) }
+  let(:sync_case_priority_double) { double(Hackney::Income::SyncCasePriority) }
 
   before do
     stub_const('Hackney::Tenancy::AddActionDiaryEntry', use_case_double)
     allow(use_case_double).to receive(:new).and_return(use_case_double)
+    allow(use_case_double).to receive(:execute)
+
+    stub_const('Hackney::Income::SyncCasePriority', sync_case_priority_double)
```
richrace: I'm not a fan of this `stub_const` I'd rather this read:

`allow(Hackney::Income::SyncCasePriority).to receive(:new).and_return(...)`

spec/controllers/action_diary_controller_spec.rb
```diff
@@ -38,6 +44,17 @@
...
   end
 
+  context 'when receiving valid params to the sync case priority use case' do
+    it 'passes the correct params to the use case' do
+      expect(sync_case_priority_double).to receive(:execute)
+        .with(tenancy_ref: action_diary_params[:tenancy_ref])
+        .and_return(nil)
+        .once
+
+      patch :create, params: action_diary_params
```
richrace: Why are we `patch`ing to a `create` action??? shouldn't it be `post`?

spec/controllers/action_diary_controller_spec.rb
richrace: /cc @kjdchapman do you have thoughts on this? re: testing strategies.

spec/controllers/action_diary_controller_spec.rb
georgeschena: Is it a 'patch' because it updates part of the resource and not a 'post' because it creates a new resource?

spec/controllers/action_diary_controller_spec.rb
kjdchapman: Overriding a constant (especially one we own) is generally a design smell, I think. 

Both versions are doing that, but I think the `allow(X).to receive...` is the more readable of the two. If we have to do this, I'd prefer it.

app/controllers/action_diary_controller.rb
georgeschena: I feel like by doing this were mixing multiple functionality into a single use case, and CA usually suggests a use case does one thing and one thing well. Personally for me and for new devs who may come onto this project it makes sense to keep all this logic separated into separate use cases.

spec/controllers/action_diary_controller_spec.rb
georgeschena: I like Richards suggestion, seems more condensed and readable imo

app/controllers/action_diary_controller.rb
richrace: Spoke about this in person. Went through pros/cons, explained in more details my views.

lib/use_cases/add_action_diary_and_sync_case.rb
```diff
@@ -0,0 +1,20 @@
+module UseCases
+  class AddActionDiaryAndSyncCase
+    def initialize(prioritisation_gateway:, sync_case_priority:, action_diary_gateway:)
+      @prioritisation_gateway = prioritisation_gateway
```
richrace: Do you need this line?

spec/use_cases/add_action_diary_and_sync_case_spec.rb
```diff
@@ -0,0 +1,32 @@
+require 'rails_helper'
+
+describe UseCases::AddActionDiaryAndSyncCase do
+  let(:add_action_diary_and_sync_case) {
+    described_class.new(prioritisation_gateway: prioritisation_gateway,
+                        sync_case_priority: sync_case_priority,
+                        action_diary_gateway: action_diary_gateway)
+  }
+
+  let(:prioritisation_gateway) { spy }
```
richrace: We never use this spy.

PR #251, opened 2019-11-14
Write to action diary and update status when downloading a letter

maysakanoni:
**What?**

- Write to the action diary when a letter before action has been downloaded

**Why?**

- So that the download action is updated on the action diary


app/controllers/documents_controller.rb
```diff
@@ -3,9 +3,26 @@ def download
...
     send_file doc_download[:filepath], type: doc.mime_type, filename: letter_file_name(doc)
   end
 
+  # def mark_downloaded
+  #   document = Hackney::Cloud::Document.find(params.fetch(:id))
+  #   if document.status != 'downloaded'
+  #     document.update(status:'downloaded')
+  #   end
+  #   head :ok
+  # end
```
kjdchapman: I'd take this out of this PR - think it'll confuse things.

app/controllers/documents_controller.rb
```diff
@@ -19,4 +36,11 @@ def letter_file_name(doc)
     letter_template = meta.dig(:template_id)
     pay_ref.to_s + '_' + letter_template.to_s + doc.extension
   end
+
+  def get_tenancy_ref(doc)
```
kjdchapman: I like pulling this out into a method.

spec/controllers/documents_controller_spec.rb
```diff
@@ -10,13 +10,30 @@
...
                                .and_return(filepath: Tempfile.new.path, document: document)
-        get :download, params: { id: document.id }
+        expect_any_instance_of(add_action_diary_use_case)
+          .to receive(:execute).with(
+            tenancy_ref: tenancy_ref,
+            action_code: 'SLB',
+            comment: 'LBA sent (SC)',
+            username: username
+          ).and_return(header: 200)
+        get :download, params: { id: document.id, username: username }
```
kjdchapman: I guess this was the failing spec?

I'm not sure if I would have fixed it - I don't like testing internals like this. 

I'd want to understand what this spec is trying to cover - if we've covered it in our new acceptance specs anyway, I'd delete this one. 

I don't know if that's the case here without getting my head around it, though!

spec/requests/downloading_a_letter_pdf_spec.rb
```diff
@@ -7,28 +7,67 @@
   let(:payment_ref) { Faker::Number.number(6) }
   let(:house_ref) { Faker::Number.number(6) }
   let(:prop_ref) { Faker::Number.number(6) }
+  let(:tenancy_ref) { Faker::Number.number(6) }
   let(:postcode) { Faker::Address.postcode }
-  let(:username) { Faker::Name.name }
   let(:email) { Faker::Internet.email }
 
   before do
+    Timecop.freeze
```
kjdchapman: :policeman:

app/controllers/documents_controller.rb
```diff
@@ -3,9 +3,26 @@ def download
     doc_download = letter_use_case_factory.download.execute(id: params.fetch(:id))
     doc = doc_download[:document]
 
+    tenancy_ref = get_tenancy_ref(doc)
+    if params[:username] != ''
```
kjdchapman: I like using the Rails methods ([`blank`](https://apidock.com/rails/Object/blank%3F) and [`present`](https://apidock.com/rails/Object/present%3F)) to make these sorts of conditionals more readable:
```suggestion
    if params[:username].blank?
```

richrace: `!= ''` is different logic than `#present?`

With `!= ''`
- Whitespace(`' '`) is true
- `nil` is true 
- `'string'` is true
- empty string is false

With `#present?` 
-  whitespace is false 
- `nil` is fale
- `'string'` is true
- empty string is false

kjdchapman: This is true - I was assuming the intent was to get behaviour like `present?`, but that might not be the case!

spec/controllers/documents_controller_spec.rb
maysakanoni: Yes this was in the controller spec, the tests failed when I added more logic to the download controller. Im interested as to why this is bad practice. In our request spec, were testing (using webmock) that hitting this controller creates a post request to the relevant api endpoint- is this sufficient enough to delete this test?

app/controllers/documents_controller.rb
maysakanoni: removed

app/controllers/documents_controller.rb
maysakanoni: Yes, ```present?``` works as intended 😄

spec/controllers/documents_controller_spec.rb
kjdchapman: Let's talk in-person about why I think testing internals is bad?

I think if the two specs are doing the same we can remove the controller one - I need to read into it a bit to understan though.

kjdchapman: Alright, had the chance to read this properly now - I think it's fine to get rid of the `describe '#download'` block in this controller spec, as it's covered in `downloading_a_letter_pdf_spec`. 

I'd keep the `describe '#index'` block until we're covering that elsewhere, though.

What do you think?

app/controllers/documents_controller.rb
```diff
@@ -3,6 +3,17 @@ def download
...
+    document_status = Hackney::Cloud::Document.find_by!(id: params[:id]).status
+    if params[:username].present? && document_status == 'uploaded' && params[:documents_view] != 'true'
+      income_use_case_factory.add_action_diary.execute(
+        tenancy_ref: tenancy_ref,
+        action_code: 'SLB',
+        comment: 'LBA sent (SC)',
+        username: params[:username]
+      )
+      Hackney::Cloud::Document.find_by!(id: params[:id]).update(status: 'downloaded')
+    end
```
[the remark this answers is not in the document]
maysakanoni: Thanks Kyle, Ive refactored this. I should have done it as a separate commit but you can check it here https://github.com/LBHackney-IT/lbh-income-api/pull/251/files#diff-a00af01be68875483d702885663e7efaR6-R18

spec/controllers/documents_controller_spec.rb
maysakanoni: Yeah, I agree. Ive done this here https://github.com/LBHackney-IT/lbh-income-api/pull/251/commits/0f418eccb342dbd9efb381ee0d44ea9f0bb2072d

app/controllers/documents_controller.rb
```diff
@@ -3,6 +3,17 @@ def download
     doc_download = letter_use_case_factory.download.execute(id: params.fetch(:id))
     doc = doc_download[:document]
 
+    tenancy_ref = get_tenancy_ref(doc)
+    document_status = Hackney::Cloud::Document.find_by!(id: params[:id]).status
```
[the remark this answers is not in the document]
maysakanoni: Thanks, I've made this change now https://github.com/LBHackney-IT/lbh-income-api/pull/251/files#diff-a00af01be68875483d702885663e7efaR6-R18

PR #252, opened 2019-11-15
Allow running all services locally

richrace:
### Why

- Allow us to run all services locally

### What 

- Add docker service to allow running from another docker compose file

### DIFF
--- a/lib/hackney/income/use_case_factory.rb
+++ b/lib/hackney/income/use_case_factory.rb
@@ -145,6 +145,13 @@ module Hackney
         )
       end
 
+      def uh_tenancies_gateway
+        Hackney::Income::UniversalHousingTenanciesGateway.new(
+          restrict_patches: ENV.fetch('RESTRICT_PATCHES', false),
+          patches: ENV.fetch('PERMITTED_PATCHES', '').split(',')
+        )
+      end
+
       private
 
       def cloud_storage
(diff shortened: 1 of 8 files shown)


lib/hackney/income/use_case_factory.rb
```diff
@@ -145,6 +145,13 @@ def get_failed_sms_messages
         )
       end
 
+      def uh_tenancies_gateway
+        Hackney::Income::UniversalHousingTenanciesGateway.new(
+          restrict_patches: ENV.fetch('RESTRICT_PATCHES', false),
+          patches: ENV.fetch('PERMITTED_PATCHES', '').split(',')
+        )
+      end
+
```
kjdchapman: This looks like an unrelated change?

lib/tasks/income/sync.rake
```diff
@@ -6,5 +6,16 @@ namespace :income do
...
+    desc 'manually runs the sync'
+    task :manual_sync do
+      use_case_factory = Hackney::Income::UseCaseFactory.new
+
+      tenancy_refs = use_case_factory.uh_tenancies_gateway.tenancies_in_arrears
+
+      tenancy_refs.each do |tenancy_ref|
+        use_case_factory.sync_case_priority.execute(tenancy_ref: tenancy_ref)
+      end
+    end
```
kjdchapman: This looks like an unrelated change, too?

bjpirt: Agreed - I think they are unrelated, although they do relate to a rake task that manually syncs (although ti isn't included in this PR so these should probably live with that)

richrace: Would you like me to remove this from this PR?

kjdchapman: I'd prefer that, if there's another good place for it to sit. 

The commit it's associated with is "rake task WIP", which doesn't tell me anything about what's going on here/why!

richrace: 👍 I'll create a separate PR.

PR #255, opened 2019-11-22
Add Income Collection Letters 1 and 2

richrace:
### Why

- As an Income Collection user I need to send Letters 1 and 2 to tenants.

### What

- Lots of unpicking of Leasehold services specific Letter generation
- Support Payment Ref and Tenancy Ref as options for finding data for letters

### Guidance for Review

- Try not worry about the architecture we will be able to address this in the future.
- Any naming changes etc would be welcome
- Tried to keep consistencies with the Leasehold Services letter generation
- Not worrying about DRYing at this moment.


lib/use_cases/generate_and_store_letter.rb
```diff
@@ -2,15 +2,25 @@
...
       letter_use_case_factory = Hackney::Letter::UseCaseFactory.new
 
-      letter_data = pdf_use_case_factory.get_preview.execute(
-        payment_ref: payment_ref,
-        template_id: template_id,
-        user: user
-      )
+      income_collection_templates = %w[income_collection_letter_1 income_collection_letter_2]
+
+      if template_id.in?(income_collection_templates)
```
kjdchapman: I can't find where this `in?` method is coming from. I'm bad at google. Help?

spec/controllers/letters_controller_spec.rb
```diff
@@ -37,8 +37,6 @@
   describe '#create' do
     let(:generate_and_store_use_case_spy) { spy }
     let(:payment_ref) { Faker::Number.number(6) }
-    let(:username) { Faker::Name.name }
```
kjdchapman: Nice, I've spotted plenty of superfluous test setup hanging around the place, good to clear it out!

spec/hackney/income/jobs/send_income_collection_letter_to_gov_notify_job_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+
+describe Hackney::Income::Jobs::SendIncomeCollectionLetterToGovNotifyJob do
+  let(:document) { create(:document, filename: 'test.txt', ext_message_id: nil) }
+  let(:document_id) { document.id }
+
+  let(:message_receipt) { Hackney::Notification::Domain::NotificationReceipt.new(body: 'body', message_id: SecureRandom.uuid) }
+
+  before {
+    expect_any_instance_of(Aws::S3::Encryption::Client).to receive(:get_object).and_return(AwsClientResponse.new)
+    expect_any_instance_of(Hackney::Notification::SendPrecompiledLetter).to receive(:execute).with(
```
kjdchapman: I'm not sure what the right way to architect these jobs are, as `JobClass.perform_now` looks kind of hard to test without lots of mocking. 

Not in a rush to solve it with this PR, but interested to hear your thoughts (here or in-person).

spec/lib/hackney/notification/send_precompiled_letter_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  context 'when sending a letter' do
+    let(:subject) do
+      send_precompiled_letter.execute(
+        unique_reference: unique_reference,
+        letter_pdf: test_file
+      )
+    end
+
+    it { expect(subject).to be_a Hackney::Notification::Domain::NotificationReceipt }
```
kjdchapman: I've not found a lot of value in testing something in a specific class. Tends to break on me when I don't want it to, and vice versa.

(it may be copied from elsewhere, but I'd be tempted to just drop it entirely)

lib/hackney/income/use_case_factory.rb
```diff
@@ -56,6 +56,12 @@ def send_precompiled_letter
         )
       end
 
+      def send_precompiled_letter_to_gov_notify
```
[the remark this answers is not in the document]
richrace: Yes. Naming is not great. However, with this method name we know all the side effects.

lib/hackney/pdf/preview_generator.rb
```diff
@@ -62,7 +62,7 @@ def get_return_date
         30.days.from_now.strftime('%d %B %Y')
       end
 
-      def get_lba_date
+      def get_sending_date
```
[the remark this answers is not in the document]
richrace: /cc @alex-demetriou / @Miles-Alford

lib/use_cases/generate_and_store_letter.rb
richrace: Rails sugar 🍬https://api.rubyonrails.org/classes/Object.html#method-i-in-3F

spec/hackney/income/jobs/send_income_collection_letter_to_gov_notify_job_spec.rb
richrace: I think in person would be better.

spec/lib/hackney/notification/send_precompiled_letter_spec.rb
richrace: Think I'd like to get your thoughts on this in person too.

PR #262, opened 2019-11-26
Lba balance bug

Miles-Alford:
**WHAT**

- Updated template paths for the LBA and letter 2, this is yet to be tested

- Update `lba_balance` calculation method

**WHY**

- LBA balance was not showing up as the directories of the templates has been updated so the correct letter class wasn't not picked up in the switch case

### DIFF
--- a/app/models/hackney/service_charge/letter/before_action.rb
+++ b/app/models/hackney/service_charge/letter/before_action.rb
@@ -3,7 +3,7 @@ module Hackney
     class Letter
       class BeforeAction < Hackney::ServiceCharge::Letter
         TEMPLATE_PATHS = [
-          'lib/hackney/pdf/templates/letter_before_action.erb'
+          'lib/hackney/pdf/templates/leasehold/letter_before_action.erb'
         ].freeze
         MANDATORY_FIELDS = %i[original_lease_date date_of_current_purchase_assignment].freeze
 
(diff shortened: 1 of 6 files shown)


app/models/hackney/service_charge/letter/before_action.rb
```diff
@@ -29,7 +31,11 @@ def freehold?
         private
 
         def calculate_lba_balance(arrears_balance, money_judgement)
-          return 0 if arrears_balance.nil? || money_judgement.nil?
+          if arrears_balance.nil?
+            arrears_balance = 0
+          elsif money_judgement.nil?
+            money_judgement = 0
+          end
```
maysakanoni: is this needed? `money judgement` can be included as a mandatory field for` BeforeAction` and `arrears balance` is always validated as `total_collectable_arrears_balance` is a mandatory field in the Letter class

app/models/hackney/service_charge/letter/before_action.rb
```diff
@@ -20,6 +21,7 @@ def initialize(params)
                                           validated_params[:money_judgement]
                                         ))
           @tenure_type = validated_params[:tenure_type]
+          validate_lba_exists(@lba_balance)
```
richrace: Shouldn't this function be `validate_lba_balance_exists?` you do not need to pass the instance variable `@lba_balance`

maysakanoni: yes it should! This has been updated

richrace:
We probably want to add a test in this file file https://github.com/LBHackney-IT/lbh-income-api/blob/master/spec/requests/letters_spec.rb that tests the expected balance is within the `preview` key.

As we saw when debugging, this value was missing from it. Whilst testing each component should be helpful, testing at the Acceptance level should indicate if we've missed tests elsewhere.

PR #264, opened 2019-11-27
Fix LBA balance bug (MA-249)

Miles-Alford:
**WHAT**
• Updated template paths for the LBA and letter 2, this is yet to be tested
• Update `lba_balance` calculation method

**WHY**
• LBA balance was not showing up as the directories of the templates has been updated so the correct letter class wasn't not picked up in the switch case

### DIFF
--- a/app/models/hackney/service_charge/letter/before_action.rb
+++ b/app/models/hackney/service_charge/letter/before_action.rb
@@ -3,7 +3,7 @@ module Hackney
     class Letter
       class BeforeAction < Hackney::ServiceCharge::Letter
         TEMPLATE_PATHS = [
-          'lib/hackney/pdf/templates/letter_before_action.erb'
+          'lib/hackney/pdf/templates/leasehold/letter_before_action.erb'
         ].freeze
         MANDATORY_FIELDS = %i[original_lease_date date_of_current_purchase_assignment].freeze
 
--- a/app/models/hackney/service_charge/letter/before_action.rb
+++ b/app/models/hackney/service_charge/letter/before_action.rb
@@ -9,7 +9,7 @@ module Hackney
 
         def initialize(params)
           super(params)
-          
+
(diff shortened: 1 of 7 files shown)


app/models/hackney/service_charge/letter/before_action.rb
```diff
@@ -20,6 +21,7 @@ def initialize(params)
                                           validated_params[:money_judgement]
                                         ))
           @tenure_type = validated_params[:tenure_type]
+          validate_lba_exists(@lba_balance)
```
richrace: I thought you changed this?

Miles-Alford: We did, made an error during our rebase, fixing right now.

spec/requests/letters_spec.rb
```diff
@@ -98,6 +101,23 @@
...
+        post messages_letters_path, params: {
+          payment_ref: payment_ref,
+          template_id: 'letter_before_action',
+          tenancy_ref: tenancy_ref,
+          user: user
+        }
+
+        keys_to_ignore = %w[uuid]
+
+        json_response = JSON.parse(response.body).except(*keys_to_ignore)
```
richrace: As we are only looking at a single key in the test, we do not need to remove any fields from the JSON

spec/requests/letters_spec.rb
```diff
@@ -98,6 +101,23 @@
         expect(json_response).to eq(expected_json_response_as_hash)
       end
 
+      it 'generates an LBA with an lba_balance' do
+        lba_template_path = 'lib/hackney/pdf/templates/leasehold/letter_before_action.erb'
+        expect(Pathname.new(lba_template_path)).to exist
```
richrace: This doesn't seem to relate to this test?

maysakanoni: You're right, we've removed this now.

PR #266, opened 2019-12-02
Ma 133 sync uc data

maysakanoni:
**What?**

- Sync in all the field related to universal credit from action diary table

**Why?**

- So we can add this to classification business rules to add a new next action, `Apply for direct payment`


lib/hackney/income/tenancy_prioritiser/universal_housing_criteria.rb
```diff
@@ -150,6 +181,22 @@ def nosp_served_date
...
+        end
+
+        def uc_direct_payment_requested
+          attributes[:uc_direct_payment_requested]
+        end
+
+        def uc_direct_payment_received
+          attributes[:uc_direct_payment_received]
+        end
+
```
kjdchapman: As a naive reader, I don't know what these fields are - is it useful to try and map the UH fields to better names?

spec/lib/hackney/income/tenancy_prioritiser/universal_housing_criteria_spec.rb
```diff
@@ -489,6 +489,62 @@
...
+    let(:date) { Date.today }
+
+    it 'can check when a rent verification code exists' do
+      create_uh_action(tenancy_ref: tenancy_ref, code: universal_credit_code, date: date)
+      expect(subject.universal_credit.instance_of?(Time)).to eq(true)
+    end
+    it 'can check when a rent verification code does not exist' do
+      expect(subject.universal_credit.instance_of?(Time)).to eq(false)
+    end
+  end
```
kjdchapman: I think the context/it blocks could do with different descriptions to help explain what the behaviour is here.

pseudocode:
```
context '#universal_credit'
  context 'when there is an action diary entry for universal credit' do
    it 'responds with a time' do
      # expectation
    end
  end

  context 'when there is an action diary entry for universal credit' do
    it 'does not respond with a time' do
      # expectation
    end
  end
end
```

spec/lib/hackney/income/tenancy_prioritiser/universal_housing_criteria_spec.rb
```diff
@@ -489,6 +489,62 @@
...
   end
 
+  describe '#universal_credit' do
+    let(:tenancy_ref) { '100000/14' }
+    let(:universal_credit_code) { 'UCC' }
+    let(:date) { Date.today }
+
+    it 'can check when a rent verification code exists' do
+      create_uh_action(tenancy_ref: tenancy_ref, code: universal_credit_code, date: date)
+      expect(subject.universal_credit.instance_of?(Time)).to eq(true)
```
kjdchapman: Checking the type of the return feels a little strange to me. 

Is the behaviour that this field will be the timestamp of when universal credit was last received? Can we check exactly that? 

e.g. `expect(subject.universal_credit).to eq(date)`

spec/lib/hackney/income/tenancy_prioritiser/universal_housing_criteria_spec.rb
```diff
@@ -489,6 +489,62 @@
...
+    end
+    it 'can check when a rent verification code does not exist' do
+      expect(subject.universal_credit.instance_of?(Time)).to eq(false)
+    end
+  end
+
+  describe '#uc_rent_verification' do
```
kjdchapman: Same comments for these as above - I'd name the it/context blocks different, and check specific output rather than types.

spec/spec_helper.rb
```diff
@@ -34,7 +34,7 @@
 
 RSpec.configure do |config|
   config.include TenancyHelper
-
+  # config.formatter = :documentation
```
kjdchapman: This looks unrelated to this PR, but it's not a big deal.

lib/hackney/income/tenancy_prioritiser/universal_housing_criteria.rb
maysakanoni: have you got any suggestions? this is something I struggled with

spec/lib/hackney/income/tenancy_prioritiser/universal_housing_criteria_spec.rb
maysakanoni: I agree, we're just checking that this code exists in the action diary, otherwise it would return nil. Im grabbing the date of entry for this code incase we need to use it in the future because returning the action code which I am looking for from the SQL query seems a bit pointless?

spec/lib/hackney/income/tenancy_prioritiser/universal_housing_criteria_spec.rb
maysakanoni: I'm happy to change this, I was trying to keep these test consistent with the other tests in the file. All the other describe blocks describe the field being returned from UH. What do you think?

spec/spec_helper.rb
maysakanoni: This was not supposed to be committed! Ive removed it

lib/hackney/income/tenancy_prioritiser/universal_housing_criteria.rb
kjdchapman: Can we talk in-person about what these fields do? I think I'll find it easier to make suggestions then.

spec/lib/hackney/income/tenancy_prioritiser/universal_housing_criteria_spec.rb
kjdchapman: Not sure I understand - if we're getting back the date, I'd check the date is correct. I think these specs would pass with an implementation like:
```
def universal_credit
  Time.now
end
```

which feels wrong to me!

spec/lib/hackney/income/tenancy_prioritiser/universal_housing_criteria_spec.rb
kjdchapman: I'm happy with the `describe` text as is - it's the `it` and `context` blocks I'd like to rename.

lib/hackney/income/tenancy_prioritiser/universal_housing_criteria.rb
elena-vi: I feel like mapping UH names to better names might introduce unnecessary confusion later down the line. In a sense that we might have a conversation and one person will your _"new better name"_ and one will use _"original UH name"_ and a mismatch of understanding might happen (I'm prone to getting confused like that btw).
That being said, I'm all for maybe having some comments in the code?

spec/lib/hackney/income/tenancy_prioritiser/universal_housing_criteria_spec.rb
maysakanoni: I've updated them now.

lib/hackney/income/tenancy_prioritiser/universal_housing_criteria.rb
kjdchapman: Yeah, I've got that worry as well @elena-vi. 

I think the strategy of naming things different works if it's one place you can refer back to, but UH's influence is kind of spread all over all of our apps at the moment, sadly.

I don't love comments for this, as you can't see them when you're a client of this class. 

I'm happy to leave this as-is in this PR - I think the job of "contain UH concepts in a particular module/component" is bigger than we have budget for right now.

PR #268, opened 2019-12-02
Sending letter one writes to action diary

elena-vi:
Since the action diary got fixed we can now write into the action diary when we send rent letters. 
I removed a bunch of nearly duplicate code here and made sending letters generic again


spec/lib/hackney/notification/send_manual_precompiled_letter_spec.rb
```diff
@@ -37,19 +37,25 @@
...
-          send_precompiled_letter.execute(
-            unique_reference: unique_reference,
-            letter_pdf: test_file,
-            payment_ref: payment_ref,
-            template_id: template
-          )
-        } .not_to raise_error
-      }
+
+    context 'when templates write into action diary they should have action codes associated with them' do
```
richrace: 👍

PR #270, opened 2019-12-03
[MA-253] Update Classification Flow

alex-demetriou:
### What
- Change Letter 2 arrears amount to 3 weeks' rent
- Remove the warning letter stage
- Change the NOSP to require letter 2 to have been sent instead of the warning letters
- Update classification rules and test

### Why
Implementation of the tenancy classification needs to be changed to reflect warning letter being removed from the flow. Other classification rules have also been updated.


app/models/hackney/income/models/case_priority.rb
```diff
@@ -4,7 +4,7 @@ module Models
       class CasePriority < ApplicationRecord
         enum classification: {
           no_action: 0, send_letter_two: 1, send_letter_one: 2, send_first_SMS: 3, send_NOSP: 4,
-          send_warning_letter: 5, apply_for_court_date: 6, send_court_warning_letter: 7
+          apply_for_court_date: 5, send_court_warning_letter: 6
```
richrace: Are we happy with these enum changes?

https://api.rubyonrails.org/v5.2.3/classes/ActiveRecord/Enum.html

If we have anything in the DB with integer value of `7`, what will it now do?

lib/hackney/income/tenancy_prioritiser/tenancy_classification.rb
```diff
@@ -113,13 +100,16 @@ def send_warning_letter?
...
+            Hackney::Tenancy::ActionCodes::INCOME_COLLECTION_LETTER_2,
+            Hackney::Tenancy::ActionCodes::INCOME_COLLECTION_LETTER_2_UH
           ]
 
           can_send_nosp = false
 
           if @criteria.nosp_expiry_date.present?
             can_send_nosp = @criteria.nosp_expiry_date < Time.zone.now
+          elsif @criteria.active_agreement?
+            can_send_nosp = false
```
richrace: Could we not just do:

```suggestion
            return false
```

app/models/hackney/income/models/case_priority.rb
elena-vi: this makes me a little uncomfortable, can we confidently say that the sync will just overwrite `7`s when it runs next?

lib/hackney/income/tenancy_prioritiser/tenancy_classification.rb
maysakanoni: Indeed, I've changed this!

app/models/hackney/income/models/case_priority.rb
richrace: I *think* we delete all records that have not been sync'd. So we *should* be fine to change the enum values.

https://github.com/LBHackney-IT/lbh-income-api/blob/master/lib/hackney/income/schedule_sync_cases.rb#L23

I'd prefer it if we do not change enum values unless there's an exceptional reason - which I do not think this is.

maysakanoni: Updated here https://github.com/LBHackney-IT/lbh-income-api/pull/270/commits/a8d1e2ef67a76f03ed76207438e62d85aabf2da3

lib/hackney/income/tenancy_prioritiser/tenancy_classification.rb
kjdchapman: Fixed this in 34a7f84ed

lib/hackney/income/tenancy_prioritiser/tenancy_classification.rb
```diff
@@ -67,43 +66,35 @@ def apply_for_court_date?
...
-            Hackney::Tenancy::ActionCodes::AUTOMATED_SMS_ACTION_CODE,
-            Hackney::Tenancy::ActionCodes::MANUAL_SMS_ACTION_CODE
+          return if @case_priority.paused?
+          return if @criteria.nosp_served?
+          return if @criteria.active_agreement?
+
+          after_letter_one_actions = [
+            Hackney::Tenancy::ActionCodes::INCOME_COLLECTION_LETTER_1,
+            Hackney::Tenancy::ActionCodes::INCOME_COLLECTION_LETTER_2,
+            Hackney::Tenancy::ActionCodes::INCOME_COLLECTION_LETTER_2_UH,
```
richrace: I like this name 👍

lib/hackney/income/tenancy_prioritiser/tenancy_classification.rb
```diff
@@ -67,43 +66,35 @@ def apply_for_court_date?
...
+            @criteria.active_agreement? == false
         end
 
         def send_letter_one?
-          valid_actions = [
-            Hackney::Tenancy::ActionCodes::AUTOMATED_SMS_ACTION_CODE,
-            Hackney::Tenancy::ActionCodes::MANUAL_SMS_ACTION_CODE
+          return if @case_priority.paused?
+          return if @criteria.nosp_served?
+          return if @criteria.active_agreement?
```
richrace: Shouldn't these `return false if` rather than `nil`?

lib/hackney/income/tenancy_prioritiser/tenancy_classification.rb
```diff
@@ -67,43 +66,35 @@ def apply_for_court_date?
...
+            Hackney::Tenancy::ActionCodes::INCOME_COLLECTION_LETTER_2_UH,
+            Hackney::Tenancy::ActionCodes::COURT_WARNING_LETTER_SENT
           ]
 
-          @criteria.last_communication_action.in?(valid_actions) &&
-            @criteria.balance > 10 &&
-            @criteria.nosp_served? == false &&
-            last_communication_between_three_months_one_week? &&
-            @case_priority.paused? == false
+          return if @criteria.last_communication_action.in?(after_letter_one_actions) &&
```
richrace:
```suggestion
          return false if @criteria.last_communication_action.in?(after_letter_one_actions) &&
```

lib/hackney/income/tenancy_prioritiser/tenancy_classification.rb
```diff
@@ -67,43 +66,35 @@ def apply_for_court_date?
...
+                    last_communication_newer_than?(3.months.ago)
+
+          @criteria.balance >= @criteria.weekly_rent
         end
 
         def send_letter_two?
           valid_actions = [
             Hackney::Tenancy::ActionCodes::INCOME_COLLECTION_LETTER_1
```
richrace: Do we not care if the Letter 1 from UH has been sent?

lib/hackney/income/tenancy_prioritiser/tenancy_classification.rb
```diff
@@ -112,35 +103,37 @@ def send_warning_letter?
         end
 
         def send_nosp?
+          return if @criteria.active_agreement?
+          return if @case_priority.paused?
```
richrace: `return false if ...`?

lib/hackney/income/tenancy_prioritiser/tenancy_classification.rb
```diff
@@ -112,35 +103,37 @@ def send_warning_letter?
...
-            @criteria.balance >= arrear_accumulation_by_number_weeks(4) &&
-            @case_priority.paused? == false
+          can_send_nosp &&
+            @criteria.nosp_served? == false &&
+            @criteria.balance >= arrear_accumulation_by_number_weeks(4)
         end
 
         def last_communication_between_three_months_one_week?
-          return false if @criteria.last_communication_date.nil?
+          return if @criteria.last_communication_date.nil?
```
richrace:
```suggestion
          return false if @criteria.last_communication_date.nil?
```

lib/hackney/income/tenancy_prioritiser/tenancy_classification.rb
```diff
@@ -112,35 +103,37 @@ def send_warning_letter?
...
+          last_communication_older_than?(1.week.ago) && last_communication_newer_than?(3.months.ago)
         end
 
-        def last_communication_date_before?(date)
+        def last_communication_older_than?(date)
           @criteria.last_communication_date <= date.to_date
         end
 
-        def last_communication_date_after?(date)
+        def last_communication_newer_than?(date)
```
richrace: 👍 Better naming for these functions. Thanks!

spec/lib/hackney/income/tenancy_prioritiser/classifications/send_letter_two_spec.rb
```diff
@@ -0,0 +1,119 @@
+require 'rails_helper'
+
+describe 'Send Letter Two Rule', type: :feature do
+  letter_1_in_arrears_sent_code = 'IC1'.freeze
```
richrace:
```suggestion
  letter_1_in_arrears_sent_code = Hackney::Tenancy::ActionCodes::INCOME_COLLECTION_LETTER_1
```

lib/hackney/income/tenancy_prioritiser/tenancy_classification.rb
kjdchapman: `action_codes` only lists a `INCOME_COLLECTION_LETTER_1`, with no UH variant. i'll ask Elaine if there's a code for this.

kjdchapman:
With a bit more thinking/reading, I heartily agree `return false` is better here - didn't twig that as I was writing these, thanks! Fixed it in https://github.com/LBHackney-IT/lbh-income-api/pull/270/commits/0485ddf1fb5be53c2c89e453939cfb0ffa8deb8d.

The condition matrices were alright - I like the extensive examples, the coverage they offer. I'm worried about their readability / size. I'd like to try some things out with them as I spend more time in this area.

lib/hackney/income/tenancy_prioritiser/tenancy_classification.rb
kjdchapman: There is, indeed, a code for this, and I've added it in in https://github.com/LBHackney-IT/lbh-income-api/pull/270/commits/55ee8d0363c92a2e8b0bb34619612f80b1e69072. Good spot!

PR #274, opened 2019-12-06
Ma 527 Automating sending letter 1 with no SMS

maysakanoni:
**What?**

- Remove requirement which checks that the last communication action was sending sms

- Add restriction which checks that the last communication action was longer than 3 month to send letter one if it falls back into this classification from a future classification 

- Remove duplicated test

**Why**

- Allow letter one to be sent even if sms has not been sent as some residents don't have their number on the system

### DIFF
--- a/app/models/hackney/income/models/case_priority.rb
+++ b/app/models/hackney/income/models/case_priority.rb
@@ -4,7 +4,7 @@ module Hackney
       class CasePriority < ApplicationRecord
         enum classification: {
           no_action: 0, send_letter_two: 1, send_letter_one: 2, send_first_SMS: 3, send_NOSP: 4,
-          send_warning_letter: 5, apply_for_court_date: 6, send_court_warning_letter: 7
+          apply_for_court_date: 6, send_court_warning_letter: 7
         }
 
         validates :case_id, presence: true, uniqueness: true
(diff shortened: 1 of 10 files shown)


app/models/hackney/income/models/case_priority.rb
```diff
@@ -4,7 +4,7 @@ module Models
       class CasePriority < ApplicationRecord
         enum classification: {
           no_action: 0, send_letter_two: 1, send_letter_one: 2, send_first_SMS: 3, send_NOSP: 4,
-          send_warning_letter: 5, apply_for_court_date: 6, send_court_warning_letter: 7
+          apply_for_court_date: 5, send_court_warning_letter: 6
```
elena-vi: this makes me a little uncomfortable, can we confidently say that the sync will just overwrite 7s when it runs next?

maysakanoni: Yeah I agree, @richrace said the same thing. What would you otherwise suggest?- can we skip a number?

kjdchapman:
> is this a duplicate of #270
Yeah, we're not sure how, but many of the changes here have ended up there.

The plan is to merge #270 and then see if any changes are still in this PR.

kjdchapman:
(Yes, these changes were entirely contained in #270)

PR #275, opened 2019-12-06
Add feature toggles for each automated rents letter

richrace:
### Why

- We want fine control over which letters we can automate

### What 

- Use env variables to allow us to turn on/off each automated letter

### Guidance for review

- I found this helper function which is quite nice:  https://github.com/LBHackney-IT/lbh-income-api/blob/master/config/feature_toggle.rb

### DIFF
--- a/spec/use_cases/case_classification_to_letter_type_map_spec.rb
+++ b/spec/use_cases/case_classification_to_letter_type_map_spec.rb
@@ -12,8 +12,10 @@ describe UseCases::CaseClassificationToLetterTypeMap do
   let(:allow_letter_two) { false }
 
   before do
-    allow(use_case).to receive(:env_allowed_to_send_letter_one?).and_return(allow_letter_one)
-    allow(use_case).to receive(:env_allowed_to_send_letter_two?).and_return(allow_letter_two)
+    allow(App::Application).to receive(:feature_toggle).with('AUTOMATE_INCOME_COLLECTION_LETTER_ONE')
+                                                       .and_return(allow_letter_one)
+    allow(App::Application).to receive(:feature_toggle).with('AUTOMATE_INCOME_COLLECTION_LETTER_TWO')
+                                                       .and_return(allow_letter_two)
   end
 
   context 'when the environment allows letter 1 to be sent and not letter 2' do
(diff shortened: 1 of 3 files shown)


spec/use_cases/case_classification_to_letter_type_map_spec.rb
```diff
@@ -1,33 +1,98 @@
...
-            tenancy_ref: Faker::Number.number(4),
-            classification: :send_letter_one)
-    }
+  let(:allow_letter_one) { false }
+  let(:allow_letter_two) { false }
 
-    it 'successfully checks the classification for letter 1' do
-      expect(subject.execute(case_priority: case_priority_letter_1)).to eq(letter_1)
+  before do
+    allow(use_case).to receive(:env_allowed_to_send_letter_one?).and_return(allow_letter_one)
```
kjdchapman: I'm worried that overriding behaviour of the unit under test could lead to some bad things.

If you're testing the unit, this means that some changes to the unit won't actually cause these tests to break. If you're testing the functionality (in ATDD-style), this means we're not closely reflecting the system as it would be used by a user.

I'd happily change environment variables for this spec - I think I remember you saying you didn't like that approach either?

richrace: Stubbing out `#env_allowed_to_send_letter_one?` is just a wrapper around a function. This function is tested independently. 

If you're looking for an improvement, I could stub out the `feature_toggle` function and ensure the correct env variable name is used?

https://github.com/LBHackney-IT/lbh-income-api/blob/master/config/feature_toggle.rb
https://github.com/LBHackney-IT/lbh-income-api/blob/master/spec/config/feature_toggle_spec.rb

I do not like stubbing out `ENV` variables within tests.

kjdchapman: I think stubbing out `feature_toggle` is better, yeah. Up to you if it's worth the effort!

richrace: It's a good job we decided to do this, as I was using the `feature_toggle` function incorrectly by the looks of things!

PR #278, opened 2019-12-09
Allow automation of letters 1 and 2 to write to the action diary

georgeschena:
### WHAT
Pass the tenancy_ref through in the flow of automation

### WHY
So it can write to the action diary once automation has sent letters 1 or 2

### DIFF
--- a/lib/hackney/notification/send_manual_precompiled_letter.rb
+++ b/lib/hackney/notification/send_manual_precompiled_letter.rb
@@ -1,7 +1,7 @@
 module Hackney
   module Notification
     class SendManualPrecompiledLetter < BaseManualGateway
-      def execute(username: nil, payment_ref: nil, template_id:, unique_reference:, letter_pdf:)
+      def execute(username: nil, payment_ref: nil, tenancy_ref:, template_id:, unique_reference:, letter_pdf:)
         send_letter_response =
           notification_gateway.send_precompiled_letter(
             unique_reference: unique_reference,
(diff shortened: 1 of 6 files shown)


lib/hackney/notification/send_manual_precompiled_letter.rb
```diff
@@ -11,7 +11,7 @@ def execute(username: nil, payment_ref: nil, template_id:, unique_reference:, le
         # FIXME: this must be in a background job => UH is unreliable
         # TODO: create job to accept exact same args as add_action_diary_usecase
 
-        tenancy_ref = leasehold_gateway.get_tenancy_ref(payment_ref: payment_ref).dig(:tenancy_ref)
+        tenancy_ref = leasehold_gateway.get_tenancy_ref(payment_ref: payment_ref).dig(:tenancy_ref) if tenancy_ref.nil?
```
richrace:
```suggestion
        tenancy_ref ||= leasehold_gateway.get_tenancy_ref(payment_ref: payment_ref).dig(:tenancy_ref)
```

More idiomatic ruby to me.

elena-vi: I forgot we can do this :)

georgeschena:
> The new tests are better. But they are still missing a final wiring test in my opinion.
> 
> We do not ensure that we're calling the `add_action_diary_usecase#execute` method with the correct variables.
> 
> https://github.com/LBHackney-IT/lbh-income-api/blob/fdc445bfd768888d6ca679aae26d4caa0ed1c684/spec/lib/hackney/notification/send_manual_precompiled_letter_spec.rb#L20
> 
> I'd like to see two tests for income/leasehold that ensure we are sending the correct params to that method.

I will do this once this is merged in as want to get this piece of work out as it has been a while. and is needed

PR #280, opened 2019-12-10
Allow send letter to take a tenancy_ref

georgeschena:
### WHAT
Allow letters controller to take a tenancy_ref

### WHY
So we can write to the action diary


spec/controllers/letters_controller_spec.rb
```diff
@@ -34,6 +37,25 @@
...
+      allow(controller).to receive(:find_document).and_return(find_document)
+    end
+
+    it 'calls the send_letter_to_gov_notify job' do
+      post :send_letter, params: {
+        uuid: uuid,
+        user: user,
+        tenancy_ref: tenancy_ref
+      }
+      expect(controller).to have_received(:send_letter_to_gov_notify)
```
richrace: Can we check that it has been received with the correct arguments?

georgeschena: Yeah, ill do that

PR #284, opened 2019-12-11
Fix: Provide payment options in income letters

alex-demetriou:
### Why
 - Payment options need to be included with Income Collection letters 1 and 2

### What
 - Define the Income Collections Payment Options template in the Preview Generator
 - Change the reference to the payment options template in the letters' templates


lib/hackney/pdf/preview_generator.rb
```diff
@@ -5,6 +5,7 @@ class PreviewGenerator
       SENDER_ADDRESS_PATH = 'lib/hackney/pdf/templates/layouts/sender_address.erb'.freeze
       SENDING_DATE_PATH = 'lib/hackney/pdf/templates/layouts/sending_date.erb'.freeze
       PAYMENT_OPTIONS_PATH = 'lib/hackney/pdf/templates/layouts/payment_options.erb'.freeze
+      INCOME_COLLECTION_PAYMENT_OPTIONS_PATH = 'lib/hackney/pdf/templates/income/partials/payment_options.html.erb'.freeze
```
richrace: This is the wrong file.

https://github.com/LBHackney-IT/lbh-income-api/blob/master/lib/hackney/pdf/income_preview_generator.rb

PR #286, opened 2019-12-13
Refactor classification methods

kjdchapman:
I've been looking at letter one/two/nosp classifications, and refactored some of these methods to try and get more clarity on the gaps (#282 and #283). 

This PR covers the rest of these methods. See individual commits for more detail about the steps I've taken.

I want to think more about how to clearly express "this policy applies to this subset of classifications", so I've pulled out the `paused?` check, but not `active_agreement?`.

I've also not made any significant changes to the test suite - I would like to try and make our examples more clear if we continue working on classifications.

### DIFF
--- a/lib/hackney/income/tenancy_prioritiser/tenancy_classification.rb
+++ b/lib/hackney/income/tenancy_prioritiser/tenancy_classification.rb
@@ -86,7 +86,7 @@ module Hackney
           return false if @criteria.last_communication_action.in?(after_letter_one_actions) &&
                           last_communication_newer_than?(3.months.ago)
 
-          @criteria.balance >= @criteria.weekly_rent
+          @criteria.balance >= arrear_accumulation_by_number_weeks(1)
         end
 
         def send_letter_two?
@@ -104,7 +104,7 @@ module Hackney
           return false if last_communication_newer_than?(1.week.ago)
           return false if last_communication_older_than?(3.months.ago)
 
-          @criteria.balance >= @criteria.weekly_rent * 3
+          @criteria.balance >= arrear_accumulation_by_number_weeks(3)
         end
 
         def send_nosp?
--- a/lib/hackney/income/tenancy_prioritiser/tenancy_classification.rb
+++ b/lib/hackney/income/tenancy_prioritiser/tenancy_classification.rb
@@ -35,14 +35,16 @@ module Hackney
         end
 
         def send_court_warning_letter?
-          return false if @criteria.nosp_served_date.blank?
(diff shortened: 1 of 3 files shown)


spec/support/stubs/stub_criteria.rb
```diff
@@ -45,7 +45,7 @@ def weekly_rent
     end
 
     def nosp_served_date
-      attributes[:nosp_served_date] || '2019-12-13 12:43:10'.to_date
+      attributes[:nosp_served_date]
```
kjdchapman: This was causing a test to produce a false positive. I'd quite like to get rid of all of these defaults.

lib/hackney/income/tenancy_prioritiser/tenancy_classification.rb
```diff
@@ -128,10 +98,27 @@ def send_nosp?
           @criteria.balance >= arrear_accumulation_by_number_weeks(4)
         end
 
-        def last_communication_between_three_months_one_week?
-          return false if @criteria.last_communication_date.nil?
+        def send_court_warning_letter?
+          return false if @criteria.active_agreement?
+
+          return false if @criteria.last_communication_action == Hackney::Tenancy::ActionCodes::COURT_WARNING_LETTER_SENT
```
maysakanoni: should we be checking here that the last communicated action was to send a nosp?

lib/hackney/income/tenancy_prioritiser/tenancy_classification.rb
```diff
@@ -128,10 +98,27 @@ def send_nosp?
...
+          return false unless @criteria.nosp_served?
+          return false if @criteria.nosp_served_date.blank?
+          return false if @criteria.nosp_served_date > 28.days.ago.to_date
 
-          last_communication_older_than?(1.week.ago) && last_communication_newer_than?(3.months.ago)
+          @criteria.balance >= arrear_accumulation_by_number_weeks(4)
+        end
+
+        def apply_for_court_date?
+          return false unless @criteria.nosp_served?
```
maysakanoni: were not checking if there is an active agreement here, and if it exists than it should have been breeched

kjdchapman: I'm not sure what the correct behaviour is around agreements/breaches for this `apply_for_court_date` classification, but this is the same behaviour as before (as this is just a refactor). 

If there's a change to behaviour to make here, I'd prefer to do it in a separate PR - can we talk in-person and possibly add something to the backlog?

PR #287, opened 2019-12-16
MA 282  - Fix UH Criteria action code list

richrace:
### Why

We are expecting action codes to be present in the classification system but we are not filtering the UH data with them.

### What

- Have a list of action codes used for the UH Criteria filtering
- Dynamically build SQL statement based on this list
- Ensure that any Action Code used in the classification has been defined in said list


spec/lib/hackney/income/tenancy_prioritiser/tenancy_classification_spec.rb
```diff
@@ -79,4 +79,51 @@
...
+    let(:method_name) { nil }
+    let(:missing_action_codes) do
+      assign_classification.send(method_name) - Hackney::Tenancy::ActionCodes::FOR_UH_CRITERIA_SQL
+    end
+
+    describe '#after_letter_one_actions' do
+      let(:method_name) { :after_letter_one_actions }
+
+      it 'contains action codes within the UH Criteria Codes' do
+        expect(missing_action_codes).to eq([])
```
kjdchapman: I understand this test suite, but it took me a moment to get my head around it. 

I think eliminating the repetition of `assign_classification.method_name` in the tests might not be worth the cost of the extra complication.

Is this a clearer way to write these tests?

```
let(:action_codes) do
  Hackney::Tenancy::ActionCodes::FOR_UH_CRITERIA_SQL
end

describe '#after_letter_one_actions' do
  let(:result) { assign_classification.after_letter_one_actions }

  it 'contains only codes on the action_codes list' do
    expect(result - action_codes).to be_empty
  end
end

File: lib/hackney/income/tenancy_prioritiser/tenancy_classification.rb [hunk 2]
```diff
@@ -132,6 +114,42 @@ def last_communication_newer_than?(date)
         def arrear_accumulation_by_number_weeks(weeks)
           @criteria.weekly_rent * weeks
         end
+
+        def after_letter_one_actions
```
kjdchapman: I hesitated before extracting these lists, as I wasn't sure they'd make things clearer. I think it's the right choice now.

lib/hackney/income/tenancy_prioritiser/universal_housing_criteria.rb
```diff
@@ -3,15 +3,156 @@ module Income
     class TenancyPrioritiser
       class UniversalHousingCriteria
         def self.for_tenancy(universal_housing_client, tenancy_ref)
-          sql = <<-SQL
+          attributes = universal_housing_client[
```
kjdchapman: It's hard to see in the diff - are there changes in this file (apart from [line 155](https://github.com/LBHackney-IT/lbh-income-api/pull/287/files#diff-266da5637de8cfb8fb901a365e8bd132R155), which I see and approve of)?

It'd help me to review this if the order changing was in a separate commit to the other changes, and the commit message for the order changing called it out as explicitly that.

lib/hackney/tenancy/action_codes.rb
```diff
@@ -26,6 +26,33 @@ module ActionCodes
...
+      MW_LETTER_ACTION_2_COMPLETED = 'ZW2'.freeze
+      MW_LBA_LETTER_COMPLETED = 'ZW3'.freeze
+      MW_LETTER_ACTION_1 = 'MW1'.freeze
+      MW_LETTER_ACTION_2 = 'MW2'.freeze
+      MW_LBA_LETTER = 'MW3'.freeze
+      TEXT_MESSAGE_SENT = 'SMS'.freeze
+      STAGE_ONE_COMPLETE = 'ZR1'.freeze
+      STAGE_THREE_COMPLETE = 'ZR3'.freeze
+
+      FOR_UH_CRITERIA_SQL = [
```
kjdchapman: Wonder if there's an expressive way to do this with metaprogramming - bit concerned that someone will miss adding a new constant to this second list.

lib/hackney/income/tenancy_prioritiser/tenancy_classification.rb
richrace: I only did this change to make my testing easier. Good to know it has made that code clearer.

lib/hackney/income/tenancy_prioritiser/universal_housing_criteria.rb
richrace: How about now?

spec/lib/hackney/income/tenancy_prioritiser/tenancy_classification_spec.rb
richrace: Played around a little and agree.

lib/hackney/tenancy/action_codes.rb
richrace: I feel like doing something like that is trying to solve a problem with a complex solution that we do not know is a real problem. I'm happy to leave this as is, see if the current test coverage is "good enough". Then if it presents as a problem, at that time we should look at more elegant/complex solutions.

lib/hackney/income/tenancy_prioritiser/universal_housing_criteria.rb
kjdchapman: Much better, thanks! These changes look good.

lib/hackney/tenancy/action_codes.rb
kjdchapman: Sure, I'm happy with that approach.

PR #289, opened 2019-12-17
DRY up court date examples

kjdchapman:
@richrace and I talked about the best way to write these specs, and agreed that trying things out was the best way to tell if it made things clearer.

This PR attempts to make these specs more readable without losing the explanatory power of examples. It also does a little work on the output of those specs, attempting to express a high-level description of each case.

I'm totally happy to throw away this change if it doesn't help the readability.
_____

Old output:

image

New output:

image

### DIFF
--- a/spec/support/shared_example_for_tenancy_classification.rb
+++ b/spec/support/shared_example_for_tenancy_classification.rb
@@ -9,7 +9,7 @@
 #  "returns `send_NOSP`"
 #
 def build_context_message(options)
-  options.each_with_object([]) do |(attribute, value), msg|
+  'when ' + options.each_with_object([]) do |(attribute, value), msg|
     next msg if attribute == :outcome
     msg << "'#{attribute}' is '#{value}'"
     msg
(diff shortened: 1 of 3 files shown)


richrace:
I like the improvements but if we enable randomised tests (which we should but that's a different issue) the ordering will not be "pretty"

spec/support/shared_example_for_tenancy_classification.rb
```diff
@@ -48,7 +48,7 @@ def build_context_message(options)
   condition_matrix.each do |options|
     message = build_context_message(options)
 
-    context "when #{message}" do
+    context(options[:description] || message) do
```
richrace: `message` will be generated for every `context` could we not do:

`context(options[:description] || build_context_message(options))`?

kjdchapman:
I'm fine with the pretty ordering being broken, I think it's still clearer what's going on for a given test even with random on. 

I definitely support us turning on random order in a separate change.

Do you think it's valuable to merge just this in? I don't think doing all the classification examples at once is a good use of time.

spec/support/shared_example_for_tenancy_classification.rb
kjdchapman: Sure, happy to make this change (https://github.com/LBHackney-IT/lbh-income-api/pull/289/commits/58730728831437cf83b1ddda4cb0d672cd3e97b8)

PR #290, opened 2019-12-17
income collection letter template fix

elena-vi:
appending empty lines to income collection letters so they pass gov notify validation

currently some address weren't passing validation because everything was brought up by one line
Screenshot 2019-12-17 at 16 04 01

### DIFF
--- a/app/models/hackney/income_collection/letter.rb
+++ b/app/models/hackney/income_collection/letter.rb
@@ -57,7 +57,7 @@ module Hackney
         # tenant address doesn't have full 5 lines, we need to add enough breaks
         # to make up for that otherwise the whole letter won't pass gov notify validation
         (5 - address_lines.length).times do
-          address_lines.push('<br>')
+          address_lines.push(' ')
         end
 
         address_lines.join('<br>')
--- a/app/models/hackney/income_collection/letter.rb
+++ b/app/models/hackney/income_collection/letter.rb
@@ -8,7 +8,7 @@ module Hackney
       attr_reader :tenancy_ref, :address_line1, :address_line2,
                   :address_line3, :address_line4, :address_post_code,
                   :payment_ref, :total_collectable_arrears_balance,
-                  :title, :forename, :surname, :errors
+                  :title, :forename, :surname, :errors, :tenant_address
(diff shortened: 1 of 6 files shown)


spec/lib/hackney/pdf/test_income_template.erb
```diff
@@ -4,9 +4,5 @@
   <li>tenancy_ref: <%=  @letter.tenancy_ref %></li>
   <li>forename: <%= @letter.forename %></li>
   <li>surname: <%= @letter.surname %></li>
-  <li>address_line1: <%= @letter.address_line1 %></li>
-  <li>address_line2: <%= @letter.address_line2 %></li>
-  <li>address_line3: <%= @letter.address_line3 %></li>
-  <li>address_post_code: <%= @letter.address_post_code %></li>
-  <li>total_collectable_arrears_balance: <%= @letter.total_collectable_arrears_balance %></li>
```
kjdchapman: Have these balance lines been removed by accident? If not, I don't understand why they've gone!

richrace:
Does this only affect the rents letters and not the leasehold letters?

spec/lib/hackney/pdf/test_income_template.erb
elena-vi: the address lines were removed on purpose, the balance was an accident

elena-vi:
yes, only rent letters - leasehold letter addresses aren't formatted this nicely

app/models/hackney/income_collection/letter.rb
```diff
@@ -45,6 +47,21 @@ def validate_mandatory_fields(mandatory_fields, letter_params)
 
         letter_params
       end
+
+      private
+
+      def build_tenant_address(address_lines)
+        address_lines = address_lines.select(&:present?)
+
+        # Our templates are designed to have 5 lines in the address window. If the
```
kjdchapman: Do we still want this comment now we've got the explanatory testing?

richrace: I think we should still have it here. Make it as obvious as possible the reasoning.

PR #291, opened 2019-12-18
[MA-280] Re-sync cases on Action Diary write

alex-demetriou:
### Why
* Cases should not persist in the worktray after their action has been performed.

### What
* Swapped out add_action_diary_entry Use Case for add_action_diary_and_sync case Use Case.


spec/features/letter_automation/syncing_spec.rb
```diff
@@ -0,0 +1,195 @@
...
+    stub_request(
+      :post,
+      'http://example.com/api/v2/tenancies/arrears-action-diary'
+    ).to_return(lambda do |_request|
+      # Mock the behaviour of the API by writing directly to UH
+      create_uh_action(tenancy_ref: tenancy_ref, code: 'IC1', date: Time.zone.now)
+
+      { status: 200, body: '', headers: {} }
+    end)
+  end
```
richrace: We might want to extract this out later and for it to be more of a mock.

E.g. using the request to write the correct tenancy ref, code, comment etc to the UH Sim

spec/features/letter_automation/syncing_spec.rb
```diff
@@ -0,0 +1,195 @@
...
+
+  def create_valid_uh_records_for_an_income_letter
```
richrace: We're using this in a few places now. 

We should think about extracting this to be a factory.

spec/features/letter_automation/syncing_spec.rb
```diff
@@ -0,0 +1,195 @@
...
+  end
+
+  def set_other_balances_to_zero
+    # UH Sim *may* have some data in it. Make sure any that is in it does
+    # not trigger the Sync process.
+    Hackney::UniversalHousing::Client.connection[:tenagree]
+                                     .where(Sequel[:tenagree][:cur_bal] > 0)
+                                     .exclude(tag_ref: tenancy_ref)
+                                     .update(cur_bal: 0)
+  end
```
richrace: This is a symptom of the UH Sim not being cleared between test runs/docker instances.

elena-vi: I've never experienced similar issues, but sounds like it. 
I'd feel more comfortable figuring out why the UH sim might still have some data and making sure everything is cleaned out between tests runs and all that. This method feels a but more like a dirty fix to me....

spec/requests/manually_sending_a_letter_syncs_spec.rb
```diff
@@ -0,0 +1,155 @@
...
+      house_ref: house_ref,
+      current_balance: current_balance
+    )
+    create_uh_househ(
+      house_ref: house_ref,
+      prop_ref: property_ref,
+      corr_preamble: 'Test',
+      corr_desig: 'Test',
+      corr_postcode: postcode,
+      house_desc: 'Test Name'
```
elena-vi: So address in UH are stored in quite an odd way, I think it would be a good idea to use values that look a bit more real in our tests

spec/support/universal_housing_helper.rb
```diff
@@ -95,12 +95,13 @@ def create_uh_arrears_agreement(tenancy_ref:, status:)
...
   def create_uh_action(tenancy_ref:, code:, date:)
-    Hackney::UniversalHousing::Client.connection[:araction].insert(
+    table = Hackney::UniversalHousing::Client.connection[:araction]
+    table.insert(
       tag_ref: tenancy_ref,
       action_code: code,
       action_date: date,
       action_set: 1,
-      action_no: 1,
+      action_no: (table.max(:action_no) || 0) + 1,
```
elena-vi: nice 👍

spec/requests/manually_sending_a_letter_syncs_spec.rb
alex-demetriou: Made some changes to use more realistic values

spec/features/letter_automation/syncing_spec.rb
richrace: Correct.

richrace: Looking at https://github.com/LBHackney-IT/lbh-income-api/pull/296/files#diff-a292f879f685751a781eafccb80f2502R36

I think we should just have that hook everywhere rather than opting in?

richrace:
> I found this change a little difficult to follow, and I have some thoughts on some of the tests.
> 
> The naming changes - from `add_action_diary_usecase` to `add_action_diary_and_sync_case_usecase` makes the surface area of this change pretty high.
> 
> If I write a use case that ends up with `do_x_and_y` as the most descriptive name, I tend to split it into `do_x` and `do_y` as it's easier to read. I combine use cases when there's a high-level use case that ties them together that's super-obvious, e.g I might wrap these steps in `make_coffee`.
> 
> ```ruby
> boil_kettle
> pour_water
> stir_in_coffee
> add_milk
> ```

I think one of the issues is, that we cannot just add to the action diary and *not* sync the case. These two actions are pretty important. For example, if we add an action diary entry but do not sync the case then it can be in the wrong bucket. I agree that if there was a clear and obvious name we could use that. But in this instance, every time we add an action diary we would *always* have to remember to sync the case.

I feel that having `do_x_and_y` is the better solution rather than:
- Having to always remember that whenever we write `do_x` we also need to do `do_y`.
- Calling it `do_x` and it also does `do_y` but without it being super obvious. 

I'm happy to rename `add_action_diary_and_sync_case_usecase` but I'm not sure what the best, clear name would be.

kjdchapman:
Mm, this is a tricky one, I can see both sides. I'm happy for us to keep it as-is if you think that's best.

spec/requests/manually_sending_a_letter_syncs_spec.rb
```diff
@@ -0,0 +1,155 @@
...
+    {
+      name: Faker::Name.name,
+      email: Faker::Internet.email,
+      groups: %w[leasehold-group income-group]
+    }
+  }
+  let!(:original_env_can_auto_letters) { ENV['CAN_AUTOMATE_LETTERS'] }
+  let!(:original_env_patch_codes_letters) { ENV['PATCH_CODES_FOR_LETTER_AUTOMATION'] }
+  let!(:original_env_can_auto_letter_one) { ENV['AUTOMATE_INCOME_COLLECTION_LETTER_ONE'] }
+  # set up a case that needs a letter one sent
```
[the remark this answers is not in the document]
alex-demetriou: That comment seemed misplaced/unnecessary so I've deleted it

spec/features/letter_automation/syncing_spec.rb
```diff
@@ -0,0 +1,195 @@
...
+  def then_a_document_is_queued
+    document = Hackney::Cloud::Document.last
+    expect(JSON.parse(document.metadata)['payment_ref']).to eq(payment_ref)
+    expect(document).to be_queued
+  end
+
+  def mock_gov_notify_client
+    stub_const('Notifications::Client', gov_notify_client)
+    allow(gov_notify_client).to receive(:new).and_return(gov_notify_client)
+    allow(gov_notify_client).to receive(:send_precompiled_letter).and_return(fake_response)
```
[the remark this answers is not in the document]
alex-demetriou: Agreed

spec/features/letter_automation/syncing_spec.rb
alex-demetriou: > I think we should just have that hook everywhere rather than opting in?
How would we set that more globally than putting it in each context?

spec/features/letter_automation/syncing_spec.rb
```diff
@@ -0,0 +1,195 @@
...
+    ENV['PATCH_CODES_FOR_LETTER_AUTOMATION'] = 'W02'
+  end
+
+  after do
+    ENV['CAN_AUTOMATE_LETTERS'] = original_env_can_auto_letters
+    ENV['PATCH_CODES_FOR_LETTER_AUTOMATION'] = original_env_patch_codes_letters
+    ENV['AUTOMATE_INCOME_COLLECTION_LETTER_ONE'] = original_env_can_auto_letter_one
+  end
+
+  context 'when the cron job runs' do
```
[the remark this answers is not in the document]
alex-demetriou: >This is a tough one to get right (and explain!), but I wonder if there's any way to move this spec structure around to make things clearer, without needing significant work.

I've moved the `it` block in `context 'when the cron job runs'` to the bottom, so that all of the nested `context`s appear before it; and all of the `before`s are at the top of their `context`, ahead of the `it`s.
Other than that, I think I'm missing what you're getting at a bit. Would it be clearer if all `it`s were inside a *nested* `context`, rather than there being a 'floating' one?

spec/features/letter_automation/syncing_spec.rb
```diff
@@ -0,0 +1,195 @@
...
+      prop_ref: property_ref,
+      sc_leasedate: leasedate
+    )
+  end
+
+  def given_a_case_exists
+    income_use_case_factory.sync_case_priority.execute(tenancy_ref: tenancy_ref)
+  end
+
+  def when_the_sync_runs(document_count_changes_by:, case_priority_count_changes_by:)
```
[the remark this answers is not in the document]
alex-demetriou: >This also looks like a then rather than a when to me, since it's checking some output.

This method is a bit of a confusing fit for me. I preferred the `when` because, while the method does assert that the counts of documents and case priorities change by some specified amount when the sync runs (suggesting that it should be a `then`), it also performs some action. The other `then`s *only* have assertions.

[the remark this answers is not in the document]
alex-demetriou: I also found the need to use it as a `then` so that the "given; when; then" flow read nicely; so there weren't `when`s after `then`s.

spec/features/letter_automation/syncing_spec.rb
```diff
@@ -0,0 +1,195 @@
...
+  def set_other_balances_to_zero
+    # UH Sim *may* have some data in it. Make sure any that is in it does
+    # not trigger the Sync process.
+    Hackney::UniversalHousing::Client.connection[:tenagree]
+                                     .where(Sequel[:tenagree][:cur_bal] > 0)
+                                     .exclude(tag_ref: tenancy_ref)
+                                     .update(cur_bal: 0)
+  end
+
+  alias_method :given_a_case_priority_is, :expect_case_priority_to_be
```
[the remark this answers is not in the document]
alex-demetriou: >I don't like that a given and a then method resolve to the same method, as it's surprising.

Agreed. I've removed the `given_a_case_priority_is` alias and 'shuffled things around a bit'.

spec/features/letter_automation/syncing_spec.rb
alex-demetriou: >We're using this in a few places now.

Extracted this method to `UniversalHousingHelper`.

spec/features/letter_automation/syncing_spec.rb
richrace: In the `spec/rails_helper.rb` after line 36

```ruby
  config.after(:each, type: :controller) { truncate_uh_tables }
  config.after(:each, type: :request) { truncate_uh_tables }
  config.after(:each, type: :feature) { truncate_uh_tables }
  config.after(:each, universal: true)  { truncate_uh_tables }
```

https://relishapp.com/rspec/rspec-core/v/3-9/docs/hooks/filters

This might have an impact on speed. So we may want to limit it to only be on `universal: true` tagged tests (and add the tag onto any specs that require it)

spec/features/letter_automation/syncing_spec.rb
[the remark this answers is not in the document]
kjdchapman: That's a good move, but not quite what I meant - let's have a chat about it in-person (hopefully tomorrow)?

(It's not in a _terrible_ state in this PR or anything, just struck me as a useful thing to note)

spec/features/letter_automation/syncing_spec.rb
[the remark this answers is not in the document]
kjdchapman: I don't like asserting in `when`s because I think it can be surprising to a reader, but I've seen other people happily do it. 

I suspect this is because of my preferences around assertions "midway through a test", rather than any given/when/then specifics.

I would tend to write multiple scenarios if there's "midway" steps I want to test as well:
```ruby
describe 'some component' do
  describe 'when an initial thing has been done' do
    before { do_initial_things }
    
    it 'behaves in some way' { }
    
    describe 'and a second thing has been done' do
      before { do_more_things }
      
      it 'behaves in some other way' { }
    end
  end
end
```

Again, this is a little hard to explain in text, happy to talk about it in-person.

spec/features/letter_automation/syncing_spec.rb
alex-demetriou: `config.include UniversalHousingHelper, universal: true`

what does this do?

alex-demetriou: >In the spec/rails_helper.rb after line 36

I've done this now - tests only took a couple of seconds longer to run.

spec/rails_helper.rb
```diff
@@ -33,6 +33,10 @@
 RSpec.configure do |config|
   config.include UniversalHousingHelper, type: :controller
   config.include UniversalHousingHelper, type: :request
+  config.include UniversalHousingHelper, type: :feature
+  config.after(:each, type: :controller) { truncate_uh_tables }
+  config.after(:each, type: :request) { truncate_uh_tables }
+  config.after(:each, type: :feature) { truncate_uh_tables }
```
richrace: Missing `config.after(:each, type: :universal) { truncate_uh_tables }`

spec/requests/manually_sending_a_letter_syncs_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+    ENV['AUTOMATE_INCOME_COLLECTION_LETTER_ONE'] = original_env_can_auto_letter_one
+  end
+
+  describe 'POST /api/v1/messages/letters/send' do
+    let(:income_use_case_factory) { Hackney::Income::UseCaseFactory.new }
+    let(:case_priority) { Hackney::Income::Models::CasePriority.last }
+
+    context 'with existing income collection letter' do
+      let(:income_collection_letter) {
+        generate_and_store_letter(tenancy_ref: tenancy_ref, template_id: template_id, user: user)
```
georgeschena: Make it `generate_and_store_income_collection_letter`

alex-demetriou: done this

PR #294, opened 2019-12-18
MA-243 Add breached court agreement flow

maysakanoni:
**What?**
- Add classification for court_ordered_agreement_breached

- There are a few assumptions made with regards to the pr because there is no field which differentiates between a court ordered agreement and an informal agreement:

1. An agreement which is entered after the court date is always a court ordered agreement
2. As long as the court date is before the entry agreement date it can be said with sufficient confidence that the agreement is court ordered, even if the entry of the agreement entry is more than a month or more.
3. Whenever the agreement status `arag_status` is changed then the field `arag_statusdate` is also updated to reflect this change.

**Why?**
- So that if a court ordered agreement is breached, then a letter can be sent to the tenant informing them of this


lib/hackney/income/tenancy_prioritiser/tenancy_classification.rb
```diff
@@ -35,6 +36,16 @@ def validate_wanted_action(wanted_action)
           raise ArgumentError, "Tried to classify a case as #{wanted_action}, but this is not on the list of valid classifications."
         end
 
+        def send_court_agreement_breach_letter?
+          return false if @criteria.number_of_broken_agreements < 1
+          return false if @criteria.active_agreement? == true
+          return false if @criteria.latest_active_agreement_date <= @criteria.courtdate
+          return false if @criteria.breach_agreement_date + 3.days > Date.today
+          return false unless @criteria.court_outcome == 'AGR'
+          return false if @criteria.last_communication_action != Hackney::Tenancy::ActionCodes::COURT_WARNING_LETTER_SENT
```
georgeschena: I would inverse this to not use `!=` as it is slightly less readable

spec/lib/hackney/income/tenancy_prioritiser/classifications/send_court_agreement_breach_letter_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+    #   outcome: :no_action,
+    #   active_agreement: true,
+    #   court_outcome: 'AGR',
+    #   courtdate: 8.days.ago,
+    #   latest_active_agreement_date: 7.days.ago,
+    #   breach_agreement_date: 1.day.ago,
+    #   last_communication_action: 'ANYTHING',
+    #   number_of_broken_agreements: 1
+
+    # }
```
georgeschena: How come this has been commented out but left in the spec?

maysakanoni: I wanted to see what people had to say about this test case, is it a valid test case. Will there be a situation where `last communicated action was not send court warning letter` before a court ordered agreement has been breached? At the moment this test doesn't pass because it's falling into the letter one bucket.

lib/hackney/income/tenancy_prioritiser/tenancy_classification.rb
```diff
@@ -150,6 +162,12 @@ def valid_actions_for_apply_for_court_date_to_progress
             Hackney::Tenancy::ActionCodes::COURT_WARNING_LETTER_SENT
           ]
         end
+
+        def valid_actions_for_court_agreement_breach_letter_to_progress
+          [
+            Hackney::Tenancy::ActionCodes::COURT_WARNING_LETTER_SENT
```
elena-vi: is this already set in the codebase?

maysakanoni: these codes exist in the action `Hackney::Tenancy::ActionCodes`

elena-vi: cool, just wanted to double check.
:shipit:

PR #296, opened 2019-12-23
Rent letter address fix

elena-vi:
the first line of address was not being pulled from UH correctly.

### DIFF
--- a/lib/hackney/income/universal_housing_income_gateway.rb
+++ b/lib/hackney/income/universal_housing_income_gateway.rb
@@ -35,25 +35,25 @@ module Hackney
                 .first
 
           raise TenancyNotFoundError unless res.present?
-          # correspondence_address1: corr_address[:preamble]&.strip,
-          # correspondence_address2: "#{corr_address[:desig]&.strip} #{corr_address[:aline1]&.strip}",
-          # correspondence_address3: corr_address[:aline2]&.strip || '',
-          # correspondence_address4: corr_address[:aline3]&.strip || '',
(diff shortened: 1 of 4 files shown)


lib/hackney/income/universal_housing_income_gateway.rb
```diff
@@ -36,18 +36,26 @@ def get_income_info(tenancy_ref:)
 
           raise TenancyNotFoundError unless res.present?
 
+          # if the property is part of a council estate the flat number and building name will be in the post_preamble
```
kjdchapman: I'd try to express the information in these comments through code structure/naming, as I worry comments rot more easily.

e.g. 
```
if council_estate_property?(res)
  # snip: council estate address things
else
  # snip: regular house things
end

def council_estate_property?(res)
  res[:post_preamble]&.strip.present?
end
```

spec/lib/hackney/income/universal_housing_income_gateway_spec.rb
```diff
@@ -0,0 +1,97 @@
+require 'rails_helper'
+
+describe Hackney::Income::UniversalHousingIncomeGateway, universal: true do
```
kjdchapman: We talked briefly in-person - not sure what this attribute is doing, but don't want to block the PR on its account!

richrace: It's a tag. See:

https://makandracards.com/makandra/32259-rspec-tagging-examples-and-example-groups

and 

https://github.com/LBHackney-IT/lbh-income-api/blob/master/spec/rails_helper.rb#L36

PR #303, opened 2020-01-07
Fix latest agreement date field

maysakanoni:
**What?**
- Update so that it is looking for `arag_startdate` instead of `arag_statusdate`

**Why?**
- The latest agreement date was incorrect, so the classifications were not  falling into the correct buckets


db/migrate/20200103101403_remove_old_prioritisation_from_case_priorities.rb
```diff
@@ -10,8 +10,10 @@ def up
                :days_since_last_payment_contribution,
                :payment_amount_delta_contribution,
                :payment_date_delta_contribution,
+               :number_of_broken_agreements_contribution,
                :active_agreement_contribution,
                :broken_court_order_contribution,
+               :nosp_served_contribution,
```
richrace: I don't think these are meant to be added? 

/cc @elena-vi

PR #304, opened 2020-01-07
Add court_breach_visit classification [MA-258]

georgeschena:
### WHAT
Add the `court_breach_visit` classification

### WHY
So we can filter out the cases which next Immediate action is to make a visit to the resident

[Link to prioritisation rules (income collection)](https://docs.google.com/spreadsheets/d/1zjh_ns8hhj4VvHfypSsv9FqlQm7-Ztmkkj8Lf5YD1QE/edit#gid=0)


lib/hackney/income/tenancy_prioritiser/tenancy_classification.rb
```diff
@@ -37,6 +38,14 @@ def validate_wanted_action(wanted_action)
           raise ArgumentError, "Tried to classify a case as #{wanted_action}, but this is not on the list of valid classifications."
         end
 
+        def court_breach_visit?
+          return false if @criteria.breach_agreement_date.blank?
+          return false if @criteria.breach_agreement_date + 10.days > Date.today
+          return false unless @criteria.last_communication_action.in?(court_breach_letter_actions) &&
+                              last_communication_newer_than?(3.months.ago)
+          true
```
richrace: Do you need the guard and the plain return?

If I've understood the rule correctly, this snippet:

`@criteria.last_communication_action.in?(court_breach_letter_actions) && last_communication_newer_than?(3.months.ago)` 

When it evaluates to `true` we want to return `true`? So, why do we add `return false unless`?

georgeschena: Yes you're right was going to change this today, you beat me to it.

spec/lib/hackney/income/tenancy_prioritiser/classifications/court_breach_visit_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+      outcome: :court_breach_visit,
+      nosps_in_last_year: 0,
+      weekly_rent: 5,
+      is_paused_until: nil,
+      balance: 15.0, # 3 * weekly_rent
+      active_agreement: false,
+      last_communication_action: court_breach_letter_code,
+      last_communication_date: Date.today + 21,
+      breach_agreement_date: 12.days.ago
+    },
```
richrace: Is it work refactoring these into the "new" style with the `description` key?

PR #306, opened 2020-01-08
Rents letter error handling

elena-vi:


lib/hackney/notification/request_precompiled_letter_state.rb
```diff
@@ -1,28 +1,37 @@
...
-        update_document(message_id: message_id, status: response[:status])
+        update_document(document: document, status: response[:status])
+
+        if document.income_collection? && document.failed?
+          related_case = case_priority_store.by_payment_ref(document.parsed_metadata[:payment_ref])
+
+          add_action_diary_usecase.execute(
+            tenancy_ref: related_case.tenancy_ref,
+            action_code: 'WAR',
+            comment: 'Neque porro quisquam est qui dolorem ipsum quia dolor sit amet, consectetur, adipisci velit.'
```
richrace: Just leaving a note to make sure this changes.

lib/hackney/income/tenancy_prioritiser/tenancy_classification.rb
```diff
@@ -37,6 +40,11 @@ def validate_wanted_action(wanted_action)
           raise ArgumentError, "Tried to classify a case as #{wanted_action}, but this is not on the list of valid classifications."
         end
 
+        def review_failed_letter?
+          return false if @documents.empty?
+          @documents.most_recent.failed? && @documents.most_recent.income_collection?
+        end
```
richrace: If the last document failed, and we fix the problem. The last document still 'failed'. How does it ever progress?

elena-vi: That's a very good question. 
This is blocked by the codes not being ready, so maybe tomorrow we could have a conversation with Soraya on what should happen next.

richrace: Sounds good.

elena-vi: Perhaps an idea would be to introduce another state that's something like 'failure reviewed' and allowing the user to manually set that state once the document has been reviewed and the issue had been addressed with gov notify, then the letter won't be picked up by this classification the next time the sync runs..

richrace: So we had a discussion in person.

Decided to leave this in for now as we don't have visibility over this error. We can work on how to make cases in this state progress after we've got this deployed 👍

lib/hackney/tenancy/action_codes.rb
```diff
@@ -10,6 +10,7 @@ module ActionCodes
       MANUAL_GREEN_SMS_ACTION_CODE = 'GMS'.freeze
       MANUAL_AMBER_SMS_ACTION_CODE = 'AMS'.freeze
 
+      LETTER_FAILED_VALIDATION_CODE = 'VFL'.freeze
```
richrace: This needs to be added in the Front end - https://github.com/LBHackney-IT/LBH-IncomeCollection/blob/master/lib/hackney/income/action_diary_entry_codes.rb#L3

elena-vi: yup, will add that later when I start the frontend work

PR #309, opened 2020-01-09
Fetch correct letter 2 action diary from uh criteria

richrace:
### Why

We were looking for action "Stage 2 Complete" but this rarely get's entered.

The correct code is "Stage 2" (`S02`) but this is used for many things. So we need to look for `S02` and `Policy Generated` in the action comment for when a letter is actually sent.

### What 

Some refactoring and addition to the SQL where statement in the UH Criteria


lib/hackney/income/tenancy_prioritiser/universal_housing_criteria.rb
```diff
@@ -190,20 +208,15 @@ def self.build_sql
...
+
             DECLARE @LastCommunicationDate SMALLDATETIME = (
-              SELECT TOP 1 action_date
-              FROM araction WITH (NOLOCK)
-              WHERE tag_ref = @TenancyRef
-              AND action_code IN (SELECT communication_types FROM @CommunicationTypes)
-              ORDER BY action_date DESC
+              #{build_last_communication_sql_query(column: 'action_date')}
             )
+
```
elena-vi: I like this

lib/hackney/tenancy/action_codes.rb
```diff
@@ -22,7 +22,7 @@ module ActionCodes
       INCOME_COLLECTION_LETTER_1_UH = 'S01'.freeze
 
       INCOME_COLLECTION_LETTER_2 = 'IC2'.freeze
-      INCOME_COLLECTION_LETTER_2_UH = 'ZR2'.freeze
```
elena-vi: are we sure we don't need to track this?

richrace: Discussed in person 👍

PR #312, opened 2020-01-10
DB is case sensitive so handle it

richrace:


spec/lib/hackney/income/tenancy_prioritiser/universal_housing_criteria_spec.rb
```diff
@@ -277,6 +277,14 @@
       end
     end
 
+    describe '#build_last_communication_sql_query' do
```
kjdchapman: Is there a way to test the high-level behaviour (e.g. set up a case-insensitive record and make sure the query works) instead?

richrace: So that involves changes to the UH Sim.

PR #317, opened 2020-01-14
Documents handling

elena-vi:
this allows users to mark documents as 'failure_reviewed" once the govnotify validation issues have been resolved. 

This is done by ` post '/documents/:id/review_failure'` from the front end

### DIFF
--- a/lib/hackney/cloud/storage.rb
+++ b/lib/hackney/cloud/storage.rb
@@ -44,7 +44,7 @@ module Hackney
 
       def all_documents(payment_ref: nil)
         if payment_ref.present?
-          document_model.by_payment_ref(payment_ref).order(created_at: :DESC)
+          document_model.by_payment_ref(payment_ref).exclude_uploaded.order(created_at: :DESC)
         else
           document_model.exclude_uploaded.order(created_at: :DESC)
         end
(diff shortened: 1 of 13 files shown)


lib/hackney/cloud/storage.rb
```diff
@@ -63,9 +65,18 @@ def upload(bucket_name, content, filename)
         @storage_adapter.upload(bucket_name: bucket_name, content: content, filename: filename)
       end
 
-      private
+      def update_document_status(document:, status:)
+        Rails.logger.info "Document ext_message_id #{document.ext_message_id} found with status #{status}"
+        document.status = status
```
richrace: Could we add some checking here whether the status exists in the Enum or which status are allowed to be used?

What happens if  I use a non-existing status it might default to `0` or error?

lib/hackney/cloud/storage.rb
```diff
@@ -63,9 +65,18 @@ def upload(bucket_name, content, filename)
...
+        document.save!
 
-      attr_reader :document_model
+        message = "Document has been set to #{status} - id: #{document.id}, uuid: #{document.uuid}"
```
richrace: Probably use a combination of `document.status` or `Document.classifications[document.status]`  and `status` to ensure that it's been set correctly.

PR #323, opened 2020-01-15
Balance is in arrears when over the weekly gross rent

richrace:
E.g. If I get charged my rent on Monday and I get paid on Friday, and I 
pay it then. I am not in arrears as I have paid my rent in the current 
week.

However, if I am a penny short. I am now in arrears on Monday by a 
penny.


richrace:
> is this more of a refactor pr?

It's a bug and refactor in one. Rather than changing all `>=` to be `>` I extracted it into a single method.

maysakanoni:
> > is this more of a refactor pr?
> 
> It's a bug and refactor in one. Rather than changing all `>=` to be `>` I extracted it into a single method.
> 
> I guess, I should have changed `>=` to `>` in one commit and then done the refactor in a second.

Yeah this would be good. I think I made an incorrect assumption from the description in you pr:

> However, if I am a penny short. I am now in arrears on Monday by a
penny.

This to be sounded as though this someone should not be in arrears if they're a penny short, so I was looking for this change in the pr.

richrace:
> This to be sounded as though this someone should not be in arrears if they're a penny short, so I was looking for this change in the pr.

Yea, basically weeks balance + 0.01 is in arrears.

PR #330, opened 2020-01-20
Recent Stage 3 should block Letter One

richrace:
### Why

A recent Stage 3 should be treated as valid recent communication 

### What 

Add Stage 3 action code to the list of accepted steps after a letter one.


lib/hackney/income/tenancy_prioritiser/tenancy_classification.rb
```diff
@@ -235,7 +235,8 @@ def after_letter_one_actions
             Hackney::Tenancy::ActionCodes::INCOME_COLLECTION_LETTER_2_UH_ALT_2,
             Hackney::Tenancy::ActionCodes::INCOME_COLLECTION_LETTER_2_UH_ALT_3,
             Hackney::Tenancy::ActionCodes::S0A_ALTERNATIVE_LETTER,
-            Hackney::Tenancy::ActionCodes::COURT_WARNING_LETTER_SENT
+            Hackney::Tenancy::ActionCodes::COURT_WARNING_LETTER_SENT,
+            Hackney::Tenancy::ActionCodes::STAGE_THREE_COMPLETE
```
elena-vi: I'd like to prepend any _magic UH stages_ with `UH` just so it doesn't seem like our system supports them 
```suggestion
            Hackney::Tenancy::ActionCodes::UH_STAGE_THREE_COMPLETE
```

richrace:
This might not be required as there was missing data on that case. A NOSP was served but not entered correctly.

PR #332, opened 2020-01-20
Refactor how we handle NOSPs

richrace:
### Why

There's some confusion around the language we use for NOSPs. Let's try to simplify this and keep everything in an object to help.

### What

Move NOSP related stuff into an object with it's only rules rather than having logic elsewhere.


richrace:
Waiting for clarification for the terms used here are correct before its ready.

lib/hackney/income/tenancy_prioritiser/universal_housing_criteria.rb
```diff
@@ -151,6 +149,31 @@ def most_recent_agreement
...
+          valid_until_date = nil
+          active = false
+          in_cool_off_period = false
+          valid = false
+
+          if nosp_served_date.present?
+            expires_date = nosp_served_date + 28.days
+            valid_until_date = expires_date + 52.weeks
+            active = expires_date < Time.zone.now
+            in_cool_off_period = expires_date > Time.zone.now
```
elena-vi: can we add some comments around how the NOSP works. 
Like after the NOSP serve date we leave the tenant alone for 28 days to allow them to pay the rent and how the cool_of_period works. 

Otherwise looks good

PR #335, opened 2020-01-21
[MA 313] Write to action diary when letter is not valid

richrace:
### Why

When running the automation, if there are any errors with our data we need show the User that there's been a problem.

### What 

- Pause the case and mark it as having Missing Data. 
- Write to the action diary recording this as well.


spec/use_cases/automate_sending_letters_spec.rb
```diff
@@ -101,16 +105,34 @@
...
+        errors = validation_errors[:errors].map { |error| "#{error[:name]}: #{error[:message]}" }.join('; ')
+
+        expect(set_tenancy_paused_status).to receive(:execute)
+          .with(
+            username: 'MANAGE ARREARS SYSTEM',
+            tenancy_ref: case_priority.tenancy_ref,
+            until_date: 1.week.from_now.iso8601,
+            action_code: 'RMD',
+            pause_reason: 'Missing Data',
+            pause_comment: match(/#{errors}/)
```
elena-vi: Do you mind writing out what the pause_comment will actually look like here?

PR #338, opened 2020-01-22
[MA-317] Refactor NOSP hash to object

richrace:
### Why

We have some complex logic so lets move it to be a Domain object that is easier to understand

### What

Create a domain object and reuse it where we care about NOSPs

### Future work

Once the Front End is updated to work with this new Object we should remove the fields we no longer care about on the Case Priority
- `nosp_expires_date`
- `active_nosp?`
- `nosp_served`


elena-vi:
image

PR #342, opened 2020-01-23
[MA-320] Only sync the patches we care about 

richrace:
There are some Tenancies with Patches like `WIC` that we do not care about so we should not sync them.

### DIFF
--- a/spec/lib/hackney/income/universal_housing_tenancies_gateway_spec.rb
+++ b/spec/lib/hackney/income/universal_housing_tenancies_gateway_spec.rb
@@ -121,7 +121,7 @@ describe Hackney::Income::UniversalHousingTenanciesGateway, universal: true do
           end
         end
 
-        context 'when a tenancy is does not have a patch and when one has an accepted patch' do
+        context 'when a tenancy either has no patch or has an accepted patch' do
           before do
             create_uh_tenancy_agreement(tenancy_ref: '00001/01', current_balance: 10.0, prop_ref: 'PROP1')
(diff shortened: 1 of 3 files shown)


spec/lib/hackney/income/universal_housing_tenancies_gateway_spec.rb
```diff
@@ -94,5 +94,59 @@
...
+            create_uh_tenancy_agreement(tenancy_ref: '00001/01', current_balance: 10.0, prop_ref: 'PROP1')
+            create_uh_property(property_ref: 'PROP1', patch_code: 'Z01')
+          end
+
+          it 'includes the tenancy' do
+            expect(subject).to eq(%w[00001/01])
+          end
+        end
+
+        context 'when a tenancy is does not have a patch and when one has an accepted patch' do
```
alex-demetriou:
```suggestion
        context 'when a tenancy either has no patch or has an accepted patch' do
```

spec/lib/hackney/income/universal_housing_tenancies_gateway_spec.rb
```diff
@@ -94,5 +94,59 @@
...
+            create_uh_property(property_ref: 'PROP2', patch_code: nil)
+          end
+
+          it 'includes the tenancy' do
+            expect(subject).to eq(%w[00001/01 00002/01])
+          end
+        end
+      end
+
+      context 'without a list of acceptable patches given' do
```
alex-demetriou:
```suggestion
      context 'without being given a list of acceptable patches' do
```

PR #343, opened 2020-01-24
Increase Timeout as we have seen long time deploys on AWS's end

richrace:
So looking at the latest Prod deploy, here are our times for each EC2 instance being created/started:

Box 1:
```
Created at 2020-01-23 18:10:00 +0000
Started at 2020-01-23 18:10:32 +0000
```

Box 2: 
```
Created at 2020-01-23 18:39:18 +0000
Started at 2020-01-23 18:39:50 +0000
```

Box 3:
```
Created at 2020-01-23 18:48:58 +0000
Started at 2020-01-23 18:49:11 +0000
```

Using these times I've increased the timeout from Circle CI to be 50min. Not sure why the ECS is taking that long to deploy though.


richrace:
The actual problem of our deployments taken ages is: 

> service <name> was unable to place a task because no container instance met all of its requirements. The closest matching container-instance <instance-id> is already using a port required by your task.

PR #347, opened 2020-01-28
[MA-324] Be more clever with calculating grace amount

richrace:
We were previously not considering payments made within a week (the grace period) of the grace amount (1 week gross rent). Therefore, if someone was in small arrears (less than one weeks gross rent) it will not be shown correctly.

This change removes payments made in the week period from the grace amount, allowing us to calculate the correct grace amount.

E.g. 
Start of week 
- balance = 140
- rent = 100
- payments = 0
- grace = rent - payments -> 100
- arrears = balance - grace -> 40

After one payment
- balance = 100
- rent = 100
- payments = 40
- grace = rent - payments -> 60
- arrears = balance - grace -> 40

So if later in the week  if there's another payment 
- balance = 40
- rent = 100
- payments = 40 + 60 -> 100
- grace = rent - payments -> 0
- arrears = balance - grace -> 40

### DIFF
--- a/lib/hackney/income/tenancy_prioritiser/tenancy_classification.rb
+++ b/lib/hackney/income/tenancy_prioritiser/tenancy_classification.rb
@@ -222,10 +222,10 @@ module Hackney
         end
 
         def balance_with_1_week_grace
-          @criteria.balance - calculate_grace_amount
+          @criteria.balance - calculated_grace_amount
         end
 
-        def calculate_grace_amount
+        def calculated_grace_amount
           grace_amount = @criteria.weekly_gross_rent + @criteria.total_payment_amount_in_week
 
           return 0 if grace_amount.negative?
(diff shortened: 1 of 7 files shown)


lib/hackney/income/tenancy_prioritiser/tenancy_classification.rb
```diff
@@ -222,7 +222,15 @@ def balance_is_in_arrears_by_amount?(amount)
         end
 
         def balance_with_1_week_grace
-          @criteria.balance - @criteria.weekly_gross_rent
+          @criteria.balance - calculate_grace_amount
+        end
+
+        def calculate_grace_amount
```
elena-vi:
```suggestion
        def calculated_grace_amount
```
nitpick, since this method also returns the grace_amount

PR #373, opened 2020-05-19
Use mysql in dev and test env

C-gyorfi:
### Context
In production/staging MySQL db is used by injecting the DB using the `DATABASE_URL` env var, while development and test environments are running SQLlite db. We want these env to be identical.

### Changes in this PR
- Update docker-compose so we can run MySQL `db-development` and `db-test` in a separate container and inject them using `DATABASE_URL` 

### Jira ticket 
https://hackney.atlassian.net/secure/RapidBoard.jspa?rapidView=30&projectKey=MAAP&modal=detail&selectedIssue=MAAP-117

### DIFF
--- a/docker-compose.service.yml
+++ b/docker-compose.service.yml
@@ -16,6 +16,8 @@ services:
       - GOV_NOTIFY_API_KEY=${GOV_NOTIFY_API_KEY}
       - TENANCY_API_HOST=http://tenancyapi:80
       - TENANCY_API_KEY=''
+      - MYSQL_PASSWORD=bar
+      - MYSQL_ROOT_PASSWORD=bar
     ports:
       - 3000:3000
     volumes:
@@ -26,9 +28,22 @@ services:
       - universal_housing_simulator
       - redis
       - tenancyapi
+      - db
   redis:
     image: redis:5.0.3-alpine
     command: ["redis-server", "--appendonly", "yes"]
     hostname: redis
     ports:
       - 6379:6379
+  db:
+    image: mysql:5.7
+    volumes:
+      - dev_data:/var/lib/mysql
+    restart: always
+    environment:
+      - MYSQL_PASSWORD=bar
+      - MYSQL_ROOT_PASSWORD=bar
+    ports:
+      - 3306:3306
+volumes:
+  dev_data: {}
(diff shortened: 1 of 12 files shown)


docker-compose.service.yml
```diff
@@ -16,6 +16,9 @@ services:
       - GOV_NOTIFY_API_KEY=${GOV_NOTIFY_API_KEY}
       - TENANCY_API_HOST=http://tenancyapi:80
       - TENANCY_API_KEY=''
+      - MYSQL_USERNAME=foo
```
bjpirt, line 19: I don't think you need these in here since they're in the database.yml

docker-compose.service.yml
```diff
@@ -26,9 +29,21 @@ services:
...
     hostname: redis
     ports:
       - 6379:6379
+  db:
+    image: mysql:5.7
+    volumes:
+      - ${PWD}/db/data:/var/lib/mysql
+    restart: always
+    environment:
+      - MYSQL_USERNAME=foo
```
bjpirt, line 45: I think you only need the root password now

docker-compose.yml
```diff
@@ -14,6 +14,9 @@ services:
       - AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY}
       - AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID}
       - GOV_NOTIFY_API_KEY=${GOV_NOTIFY_API_KEY}
+      - MYSQL_USERNAME=foo
```
bjpirt, line 17: Same as comment above

docker-compose.yml
```diff
@@ -32,3 +36,19 @@ services:
     hostname: redis
     ports:
       - 6379:6379
+  db:
+    image: mysql:5.7
+    volumes:
+      - dev_data:/var/lib/mysql
+    restart: always
+    environment:
+      - MYSQL_DATABASE=app-database-development
```
bjpirt, line 45: Same as comment above

docker-compose.yml
```diff
@@ -32,3 +36,19 @@ services:
...
+      - MYSQL_USERNAME=foo
+      - MYSQL_PASSWORD=bar
+      - MYSQL_ROOT_PASSWORD=bar
+    ports:
+      - "3306:3306"
```
bjpirt, line 50: No quotes needed here

PR #377, opened 2020-05-20
Add pull request template

C-gyorfi:
## Context
As suggested on https://github.com/LBHackney-IT/lbh-income-api/pull/376 by @antonyoneill 
`I think it would also be useful to agree on a standard PR template to ensure that we capture the correct information to provide context to reviewers`

This is based on the template we used on Apply and I think it worked really well. This is just a proposal and I encourage you to suggest any changes or any different template. Also, some fields are optional, so don't be pressured to fill up everything...

- Add a pull request template, so we can use a standard format when opening/reviewing a PR


antonyoneill:
I like it! Would suggest just running it past the others

PR #378, opened 2020-05-20
Working day fix

Ninamma:
### Context
Currently when previewing letters*, the date on the letters is set to tomorrow, no matter what day of the week it is. 

So if a user generates a letter on a Friday, it would have Saturday's date on it. 
Things like Bank Holidays also weren't taken into account.

*currently affects 'Letter 1/2 in arrears...'
image

### Changes in this PR
Now the preview generator will use the shiny new ✨ Bank Holidays Retriever ✨ to find the next working day to put on the letter. 

It uses the Government's [Bank Holidays API](https://www.gov.uk/bank-holidays.json) to get an array of dates - which is cached for a day to be reused for any other previews.

I had help from [this article](https://medium.com/just-tech/working-days-calculator-262e6637377f)! 🙏 

### How to test
- Pull the branch!
- Pick a Friday/day before bank holiday and use Timecop.freeze in `lib/hackney/pdf/preview_generator.rb`.
image
- `make run-all` in LBH-IncomeCollection
- Get any payment ref and send them a 'Letter 1/2 in arrears' and check the date there!

### Jira ticket 
https://hackney.atlassian.net/browse/MAAP-120?atlOrigin=eyJpIjoiZjU2ZTVhNGQ1Yjg0NGVjMmI5ZWVjY2VmZjY2OGU0MGUiLCJwIjoiaiJ9


lib/hackney/pdf/bank_holidays_retriever.rb
```diff
@@ -0,0 +1,40 @@
...
+  module PDF
+    class BankHolidaysRetriever
+      UnsuccessfulRetrievalError = Class.new(StandardError)
+      API_URL = 'https://www.gov.uk/bank-holidays.json'
+      DEFAULT_GROUP = 'england-and-wales'
+
+      def execute
+        make_request
+        begin
+          get_dates
```
antonyoneill, line 11: Not at all a comment on your code, just in general.. Code flow in Ruby is so strange.

I'm used to a more functional approach without side effects.

In JS you'd have something like
```javascript
async function execute() {
  const result = await makeRequest();
  try {
    return getDates(result);
  } catch (err) {
    return []
  }
}
```

This hurts my head :joy:

lib/hackney/pdf/bank_holidays_retriever.rb
```diff
@@ -0,0 +1,40 @@
+module Hackney
+  module PDF
+    class BankHolidaysRetriever
+      UnsuccessfulRetrievalError = Class.new(StandardError)
+      API_URL = 'https://www.gov.uk/bank-holidays.json'
+      DEFAULT_GROUP = 'england-and-wales'
+
+      def execute
+        make_request
```
antonyoneill, line 9: I don't yet understand the execution model of this project. A simple project served via Apache for instance would have one instance per request.

In other platforms an instance is shared between requests, like in Spring Boot.

Do we want to/can we consider caching this request?

I believe there is a bulk letter sending tool - what if someone tries to send 100 letters, would we make the request 100 times to the API? I wonder if we'd be rate limited, etc etc

Food for thought, I don't have the answers!!

lib/hackney/pdf/bank_holidays_retriever.rb
```diff
@@ -0,0 +1,40 @@
+module Hackney
+  module PDF
```
C-gyorfi, line 2: I wonder why is this part of the PDF module?

lib/hackney/pdf/preview_generator.rb
```diff
@@ -50,9 +50,21 @@ def execute(letter_params:, username:)
       end
 
       private
+      def get_bank_holidays
```
C-gyorfi, line 53: I saw we are using `get` in other methods and this is just a naming convention, but its a bit .Net flavoured, you could just use `bank_holidays` and I think line 65 would still work (without the assignment on line 63):
```ruby
possible_date += 1.day while bank_holidays.include?(possible_date.strftime('%Y-%m-%d')) == true
```

lib/hackney/pdf/bank_holidays_retriever.rb
elena-vi: to me it makes sense to me for this to be here because we currently only care about bank holidays while generating letters 🤷‍♀️

lib/hackney/pdf/bank_holidays_retriever.rb
Ninamma: Not sure if I still fully understand what "side effects" really means 🤔 but I think I'm starting to get it - I've reworked the code slightly, wondering if it's a bit better now? https://github.com/LBHackney-IT/lbh-income-api/pull/378/files#diff-ea3bf96b2cf5849b86abdb049062829f

lib/hackney/pdf/bank_holidays_retriever.rb
Ninamma: Added caching now! 🎉

lib/hackney/pdf/preview_generator.rb
Ninamma: Did look into this - but found that because that's in a while loop, without the assignment, we would end up hitting the cache each time to get the array of bank holidays - which isn't really necessary.

I could rename it `bank_holidays` but not sure - cause it does look quite nice with al the other `get...` functions 😅

lib/hackney/pdf/bank_holidays_retriever.rb
antonyoneill: Side effects typically describe circumstances where a function has mutated some state/variable outside of its scope. It's a functional programming thing that makes code easier to read and can eliminate bugs.

I guess it's more of a paradigm shift away from functional programming which I need to get used to.

[This article](https://medium.com/javascript-scene/master-the-javascript-interview-what-is-functional-programming-7f218c68b3a0) describes it better:

>A side effect is any application state change that is observable outside the called function other than its return value. Side effects include:
> 
>     Modifying any external variable or object property (e.g., a global variable, or a variable in the parent function scope chain)
>     Logging to the console
>     Writing to the screen
>     Writing to a file
>     Writing to the network
>     Triggering any external process
>     Calling any other functions with side-effects
> 
> Side effects are mostly avoided in functional programming, which makes the effects of a program much easier to understand, and much easier to test.
> 
> Haskell and other functional languages frequently isolate and encapsulate side effects from pure functions using monads. The topic of monads is deep enough to write a book on, so we’ll save that for later.
> 
> What you do need to know right now is that side-effect actions need to be isolated from the rest of your software. If you keep your side effects separate from the rest of your program logic, your software will be much easier to extend, refactor, debug, test, and maintain.
> 
> This is the reason that most front-end frameworks encourage users to manage state and component rendering in separate, loosely coupled modules.

lib/hackney/pdf/preview_generator.rb
```diff
@@ -51,8 +51,20 @@ def execute(letter_params:, username:)
...
-        (Time.now + 1.day).strftime('%d %B %Y')
+        possible_date = Time.now + 1.day
+        possible_date += 2.day if possible_date.saturday?
+        possible_date += 1.day if possible_date.sunday?
+
+        bank_holidays = get_bank_holidays
+
+        possible_date += 1.day while bank_holidays.include?(possible_date.strftime('%Y-%m-%d')) == true
+
+        possible_date.strftime('%d %B %Y')
```
elena-vi, line 67: radical idea: could we just move all of this logic to `BankHolidaysRetriever` and rename that calls as `NextWorkingDayRetriever`?

PR #380, opened 2020-05-21
148-NOSP actionable date

Miles-Alford:
## Context

Due to the current COVID-19 outbreak the actionable date for a NOSP has been extended from 28 days to 3 months

You can now pass an actionable date which is used to calculate the expiry date on the NOSP model. The default time for this date is 28 days so in the future all you have to do is remove the active date passed to the NOSP. Tests will be updated to reflect the change in actionable time
## Guidance to review

Happy to hear some thoughts on how to improve. Would be important to check the tests that will have to change. Ben's Idea on parameterising this so there only needs to be one place to change the actionable date is one improvement but unsure how to go about it
## Link to Jira card

https://hackney.atlassian.net/secure/RapidBoard.jspa?rapidView=30&projectKey=MAAP&modal=detail&selectedIssue=MAAP-148

### DIFF
--- a/lib/hackney/domain/nosp.rb
+++ b/lib/hackney/domain/nosp.rb
@@ -3,21 +3,22 @@ module Hackney
     # What a NOSP is:
     #
     # Expires Date is when the NOSP can now be actioned upon.
-    #   The Tenant has active_time (default 28 days) to pay the arrears before Court action can take place.
+    #   The Tenant has active_time (normally 28 days) to pay the arrears before Court action can take place.
     #   This is the "cooling off period".
     #
     # Served - This is when there is a date from UH
-    # Cool off Period - This is the period from NOSP served to active_time (default 28 days) after
+    # Cool off Period - This is the period from NOSP served to active_time (normally 28 days) after
     # Active - This when the Tenant can be taken to court if the arrears are "high" enough; after
     #          the cooling off period.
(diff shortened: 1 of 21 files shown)


lib/hackney/domain/nosp.rb
```diff
@@ -38,10 +38,10 @@ def in_cool_off_period?
 
       private
 
-      def calculate_properties
+      def calculate_properties(active_time:)
         return unless served?
 
-        @expires_date = @served_date + 28.days
+        @expires_date = @served_date + active_time
```
C-gyorfi, line 44: Maybe we could define a constant ACTIVE_TIME?

spec/controllers/tenancies_controller_spec.rb
```diff
@@ -114,10 +115,10 @@
             tenancy_ref: tenancy_1.tenancy_ref,
             nosp: {
               active: true,
-              expires_date: (nosp_served_date + 28.days).strftime('%FT%T.000Z'),
+              expires_date: (nosp_served_date + 3.months).strftime('%FT%T.000Z'),
```
C-gyorfi, line 118: And you could write these tests accordingly e.g: 
```Ruby
expires_date: (nosp_served_date + Nosp::ACTIVE_TIME).strftime('%FT%T.000Z')
```

So next time we need to change the active time, all we need to update is the constant itself

PR #381, opened 2020-05-22
Wait for DB in development/test

antonyoneill:
## Context

With a separate MySQL container, the first time it boots it can take a
little while to populate the docker volume with the necessary data.

This will happen whenever the `dev_data` docker volume is removed.

* Add `dockerize`, a helper script that waits for services to
be connectable before allowing the service to start.
* Simplify the startup commands for the service.
* Auto create db if it doesn't already exist before migration.

All the above mean that every time you start the service using `docker-compose up` you'll get a database with all the right schema (whether it existed before or not).

See:
* https://github.com/jwilder/dockerize
* https://docs.docker.com/compose/rails/

## Guidance to review

Would be good to get another set of eyes on the Dockerfile, and docker-compose files.

I'm not sure if we'll want to remove the `["sh","-c","rails db:migrate ; rails s"]` commands from the web task in AWS for staging and production. The image will now run migration as part of the entrypoint, running migrate twice may just slow boot time down a little.

## Link to Jira card

I don't have one - should I raise one first?

## Things to check

- [x] ~If this code includes a migration adding or changing columns, it also backfills existing records for consistency~
- [x] ~Environment variables have been updated~

### DIFF
--- a/bin/entrypoint.sh
+++ b/bin/entrypoint.sh
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+
+set -e
+
+echo "Starting $RAILS_ENV"
+
+if [ "$RAILS_ENV" = "test" ] || [ "$RAILS_ENV" = "development" ]
+then
+  DATABASE_SERVICE=tcp://incomeapi-db:3306
+  echo "Waiting for database to come up on $DATABASE_SERVICE"
+  # Wait for database to come up
+  dockerize -wait "$DATABASE_SERVICE" -timeout 1m
+fi
+
+# Migrate database
+rails db:migrate
+
+# Start app
+exec "$@"
--- a/bin/entrypoint.sh
+++ b/bin/entrypoint.sh
@@ -10,6 +10,9 @@ then
   echo "Waiting for database to come up on $DATABASE_SERVICE"
   # Wait for database to come up
   dockerize -wait "$DATABASE_SERVICE" -timeout 1m
+
+  # Create database - will not fail if it already exists
+  rails db:create
 fi
 
 # Migrate database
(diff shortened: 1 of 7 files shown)


bin/entrypoint.sh
```diff
@@ -0,0 +1,22 @@
...
+
+if [ "$RAILS_ENV" = "test" ] || [ "$RAILS_ENV" = "development" ]
+then
+  DATABASE_SERVICE=tcp://incomeapi-db:3306
+  echo "Waiting for database to come up on $DATABASE_SERVICE"
+  # Wait for database to come up
+  dockerize -wait "$DATABASE_SERVICE" -timeout 1m
+
+  # Create database - will not fail if it already exists
+  rails db:create
```
C-gyorfi, line 15: The problem with this is any time you are running `make test` it is trying to create a new db in dev, because the app container is always `development` env. 

To work this around I've added `make test-setup` which is:
```bash
	docker-compose run --rm app /bin/bash -c "rake db:drop RAILS_ENV=test \
																						&& rake db:create RAILS_ENV=test \
																						&& rake db:migrate RAILS_ENV=test"
```

So maybe better to keep db:create and db:migrate in `docker-compose.service.yml `

antonyoneill: Let's talk about it on Tuesday - I'm keen that `docker-compose up` is enough to make the service function correctly, it's a weird flow to have to set up a database that is defined in the compose file beforehand

antonyoneill:
Thanks for taking a look - I tested it 'from scratch' by deleting all the docker containers, and the volumes

```bash
docker kill $(docker ps -a)
docker rm $(docker ps -a -q)
docker volume rm $(docker volume ls -q)
```

PR #384, opened 2020-05-28
Refactor Classification Engine - Phase 1

antonyoneill:
## Context

We want to make it easier to extend the classification engine, in order to do so we need to refactor it so that we can make changes confidently without affecting other rules.

This is phase 1 of the refactoring process. In this PR we:
* Move the old implementation to `TenancyClassificationV1`
* Call `TenancyClassifactionV1` from `TenancyClassification` to keep the changes small
* Implement the new engine as `TenancyClassificationV2`, including contract tests to ensure the two engines provide the same actions

## Link to Jira card

[MAAP-195](https://hackney.atlassian.net/secure/RapidBoard.jspa?rapidView=30&projectKey=MAAP&modal=detail&selectedIssue=MAAP-195)
## Things to check


lib/hackney/income/tenancy_classification.rb
```diff
@@ -2,355 +2,27 @@ module Hackney
...
+          case_priority,
+          criteria,
+          documents
+        )
       end
 
       def execute
-        wanted_action = nil
+        version1_action = @version1.execute
+        version2_action = @version2.execute
```
bjpirt, line 19: I think this needs exception handling wrapping around it because if the v2 classifier has two classifications it will blow up and take the whole thing down

lib/hackney/income/tenancy_classification_v2.rb
```diff
@@ -0,0 +1,376 @@
...
+        actions << :send_NOSP if send_nosp?
+
+        actions << :send_letter_two if send_letter_two?
+        actions << :send_letter_one if send_letter_one?
+        actions << :send_first_SMS if send_sms?
+
+        actions << :no_action if actions.none?
+
+        if actions.length > 1
+          raise "Error: Multiple recommended actions #{actions} found for criteria." unless actions == %i[send_letter_one send_first_SMS]
```
bjpirt, line 39: (following on from previous comment)
alternatively this could just log erather than raising for now

antonyoneill: Ah yes, you're absolutely right. I'm open to ideas on how we can easily extract this - grepping logs may be easy enough, or sentry alerts

lib/hackney/income/tenancy_classification.rb
antonyoneill: Would prefer not to raise the other error

bjpirt:
One more thought on the general organisation of this: might be worth putting the classifiers in their own v1 and v2 folders so that we can more tidily break the v2 classifier into separate files. Could always be done in the future though.

Also, love the idea of the contract testing!

lib/hackney/income/tenancy_classification.rb
bjpirt: 👍

lib/hackney/income/tenancy_classification/classifier.rb
```diff
@@ -0,0 +1,37 @@
...
+            documents
+          )
+        end
+
+        def execute
+          version1_action = @version1_classifier.execute
+          version2_action = @version2_classifier.execute
+
+          if version1_action != version2_action
+            Rails.logger.error(
```
C-gyorfi, line 23: This one is quite interesting and I haven't seen it before, is this a common pattern to test a new version of an engine?

antonyoneill: Yeah, this is a very basic way of validating that the two classes behave in the same way, or at least providing detail as to what happened when they diverge.

There's actually a whole set of tooling around checking performance and responses from two or more different implementations, at least in PHP ( https://scientist.readme.io/docs/introduction ). I suspect there's a similar thing for Ruby/Rails but I think this simple versino will do for now

PR #387, opened 2020-06-02
Refactor review_failed_letter ruleset

antonyoneill:
## Context

This is part of the refactor around the classification rule engine for tenancies.

Seperate out the review failed letter ruleset into it's own file.

This one is a simple case which does not yet contain any special helpers, but refactoring and simplification of the helpers in a later stage may introduce some here.

## Guidance to review

Does this approach look OK?
In an ideal world, eventually each of these rulesets would be dynamically executed, rather than being explicitly included within the classifier - this would make it easy to add more in the future.

## Link to Jira card

https://hackney.atlassian.net/secure/RapidBoard.jspa?rapidView=30&projectKey=MAAP&modal=detail&selectedIssue=MAAP-201

## Things to check


antonyoneill:
Hey @Miles-Alford do you mind just checking this quickly please?

PR #391, opened 2020-06-04
Maap 206 handle failed sms

elena-vi:
## Context

When an sms fails to send because of validation errors, the API endpoint didn't send back an error

- Added another validation check (same regex as gov notify use)
- Make the api return the errors 

## Link to Jira card

https://hackney.atlassian.net/browse/MAAP-206

## Things to check

- [x] Accompanying front end PR needs to be merged in before this change: https://github.com/LBHackney-IT/LBH-IncomeCollection/pull/304


elena-vi:
Numquam est iusto v

lib/hackney/notification/gov_notify_gateway.rb
```diff
@@ -14,24 +14,24 @@ def initialize(sms_sender_id:, api_key:, send_live_communications:, test_phone_n
       end
 
       def send_text_message(phone_number:, template_id:, reference:, variables:)
-        responce = client.send_sms(
+        response = client.send_sms(
```
C-gyorfi, line 17: well spotted :D

lib/hackney/notification/send_manual_sms.rb
```diff
@@ -5,7 +5,7 @@ module Notification
     class SendManualSms < BaseManualGateway
       def execute(username:, tenancy_ref:, template_id:, phone_number:, reference:, variables:)
         phone = Phonelib.parse(phone_number)
-        if phone.valid?
+        if phone.valid? && govnotify_valid?(phone.full_e164)
```
C-gyorfi, line 8: Probably not harmful, but do we need the old `phone.valid?` check or `govnotify_valid` would be enough?

C-gyorfi:
Looks like there is a flaky test caused by MySQL, sad(ill have a look tomorrow) this is not something that introduced in this PR so good to go!

elena-vi:
I can fix the test, don't you worry

PR #393, opened 2020-06-05
Create agreements table

C-gyorfi:
## Context
We want to allow CC/LCW's to be able to add an agreement so that tenants can repay their arrears over an agreed term. This is the first slice, to allow saving an agreement in the DB

- Create `agreements` table
- Add `Agreement` model and `agreement_type` and `frequency` constraints 

## Link to Jira card

## Things to check

### DIFF
--- a/app/models/hackney/income/models/agreement.rb
+++ b/app/models/hackney/income/models/agreement.rb
@@ -1,0 +1,11 @@
+module Hackney
+  module Income
+    module Models
+      class Agreement < ApplicationRecord
+        has_one :agreement_state, class_name: 'Hackney::Income::Models::AgreementState'
+        enum agreement_type: { informal: 0, formal: 1 }
+        enum frequency: { weekly: 0, monthly: 1 }
+      end
+    end
+  end
+end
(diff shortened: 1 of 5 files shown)


app/models/hackney/income/models/agreement.rb
```diff
@@ -0,0 +1,17 @@
+module Hackney
+  module Income
+    module Models
+      class Agreement < ApplicationRecord
+        enum agreement_type: {
```
elena-vi, line 5: I'm pretty sure you can just define this as: 
```ruby
enum agreement_type: { informal: 0, formal: 1 }
```

https://api.rubyonrails.org/v5.2.3/classes/ActiveRecord/Enum.html

C-gyorfi: Yeah, I though 0 and 1 are meaningless in this case, so why not store them as a string with these defined values

elena-vi: the app will still treat them as symbols and well be able to call them like this
```ruby
Income::Models::Agreement.where(agreement_type: :formal)
```

PR #396, opened 2020-06-09
Add api-docs page with swagger integration

C-gyorfi:
## Context

We want to add clear documentation to the `income-api` so that other services can easily use it. 

- Add `/api-docs` path where we can include the API documentation going forward
- Add documentation for some of the new `agreements` endpoints 

### API docs page
image

## Guidance to review

- I've used the [rwag gem](https://github.com/rswag/rswag) to set up a documentation path and render the yaml file in a digestible format. Is there a better way to integrate swagger into our rails app?
- Have a look at the `api.yml`, does this documentation make sense? This is the first time I'm documenting an API so any suggestion would help. I've used the editor on [swagger hub](https://app.swaggerhub.com/home), but going forward we could use the `rswag-specs` gem to generate the yml file based on integration tests

## Link to Jira card

## Things to check


bjpirt:
A few comments on the API to bring it in line with Hackney standards:
- search parameters should be snake_cased - agreementType -> agreement_type (this is because we don't like case sensitive URLs)
- (do we actually need the search parameters at all? I'm not sure there's a need to search by these things)
- we should also return an id in the record
- I think we should probably also return the history in the record as an array of objects
- I think `tenancyId` should be a string because it is the UH tag_ref which is non-numeric
- should `startingBalance` be numeric?
- what is `frequency` ? Should it be an enum as string?

C-gyorfi:
> * (do we actually need the search parameters at all? I'm not sure there's a need to search by these things)

Removed them for now, if we need them later we can iterate on this.

> * I think we should probably also return the history in the record as an array of objects

Do you mean adding a history field on the `agreement` record, if so what would be a history object look like? I'm a bit confused, I thought the `/agreements/{tenancy_id}` would return an array of agreements that are associated with the tenancy and there would be maximum one `active` agreement and the rest are `inactive` 

> * I think `tenancyId` should be a string because it is the UH tag_ref which is non-numeric

I've changed that to string

> * should `startingBalance` be numeric?

We store money values as a `decimal` in db -> numeric and integer are the only accepted numeric formats in yml 

> * what is `frequency` ? Should it be an enum as string?

Currently, we store them as an integer in DB but we could return a string value in the response

bjpirt:
> Do you mean adding a history field on the agreement record?

Yes - we discussed storing the status as another table so we can see when it changed over time. e.g. 
```
{
    "history": [
         {"status": "breached", "date": "2020-04-05"},
          ...
    ],
     ....
}

bjpirt:
For `frequency` I think we should be returning a string - storing the enum as an int in the db is an optimisation and we should return user friendly responses via the API without having to look up codes

bjpirt:
Sory, spotted one more thing: the url for POST should just be `/agreements/{tenancy_id}`

bjpirt:
`tenancyId` should be a string

bjpirt:
Also, my bad - I said `status` for the history but we should make it match the top level attribute which is currently `state` - either is fine by me

C-gyorfi:
> Sory, spotted one more thing: the url for POST should just be `/agreements/{tenancy_id}`

Don't be sorry, thanks for doing it! :D

elena-vi:
> Sory, spotted one more thing: the url for POST should just be `/agreements/{tenancy_id}`

Can we refer to the `tenancy_id` as `tenancy_ref` please?

PR #398, opened 2020-06-09
Implement get agreements endpoint

C-gyorfi:
## Context
We want to allow CC/LCW's to be able to view and create existing agreements so that tenants can repay their arrears over an agreed term. 

- Add endpoint to get existing agreements for tenancy

## Guidance to review

This is a basic implementation to GET `agreements` for a given `tanancy_ref` is there any case that's missing from tests?

## Link to Jira card

## Things to check

### DIFF
--- a/app/controllers/agreements_controller.rb
+++ b/app/controllers/agreements_controller.rb
@@ -1,0 +1,17 @@
+class AgreementsController < ApplicationController
+  def index
+    requested_agreements = Hackney::Income::Models::Agreement.where(tenancy_ref: agreements_params.fetch(:tenancy_ref))
+
+    agreements = requested_agreements.map do |agreement|
+      {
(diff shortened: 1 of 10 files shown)


app/controllers/agreements_controller.rb
```diff
@@ -0,0 +1,13 @@
+class AgreementsController < ApplicationController
+  def index
+    tenancy_ref = agreements_params.fetch(:tenancy_ref)
+
+    response = Hackney::Income::ViewAgreements.execute(tenancy_ref: tenancy_ref)
```
elena-vi, line 5: Can we call this via the `income_use_case_factory` like we do in the other controllers just for consistency sake? 
```suggestion
    response = income_use_case_factory.view_agreements.execute(tenancy_ref: tenancy_ref)
```

PR #410, opened 2020-06-16
Implement create agreements endpoint

Ninamma:
## Context

We want to allow CC/LCW's to be able to view and create existing agreements so that tenants can repay their arrears over an agreed term.

- Adds CreateAgreement usecase and endpoint
- Refactors ViewAgreeements usecase
- Adds AgreementHelper that's used in both usescases listed above
- Adds constraints to the AgreementState model
- Starts using `shoulda-matchers` gem for better assertions in tests - especially for validations

## Guidance to review

The current_state of an Agreement has been refactored - is there a better way to do this? 
Any feedback on that would be good.

## Link to Jira card

## Things to check


spec/requests/agreements_spec.rb
```diff
@@ -1,53 +1,121 @@
...
-      expect(parsed_response['agreements'].first['startingBalance']).to eq(starting_balance)
-      expect(parsed_response['agreements'].first['amount']).to eq(amount)
-      expect(parsed_response['agreements'].first['startDate']).to eq(start_date.to_s)
-      expect(parsed_response['agreements'].first['frequency']).to eq(frequency)
-      expect(parsed_response['agreements'].first['currentState']).to eq(current_state)
-      expect(parsed_response['agreements'].first['history']).to eq([])
+        parsed_response = JSON.parse(response.body)
+        expect(parsed_response['tenancyRef']).to eq(tenancy_ref)
+        expect(parsed_response['agreementType']).to eq(agreement_type)
+        # TODO: starting_balance
```
C-gyorfi, line 112: I've just noticed we have some comments left in

spec/requests/agreements_spec.rb
```diff
@@ -1,53 +1,121 @@
...
+          frequency: frequency
+        }
+      end
+
+      let(:create_agreement_response) do
+        {
+          id: 1,
+          tenancyRef: tenancy_ref,
+          agreementType: agreement_type,
+          # TODO: starting_balance
```
C-gyorfi, line 83: I've just noticed we have some comments left in

lib/hackney/income/use_case_factory.rb
```diff
@@ -198,7 +198,11 @@ def uh_tenancies_gateway
       end
 
       def view_agreements
-        Hackney::Income::ViewAgreements
+        Hackney::Income::ViewAgreements.new
```
antonyoneill, line 201: What's the difference between these statements?

lib/hackney/income/agreement_helper.rb
```diff
@@ -0,0 +1,28 @@
+module Hackney
+  module Income
+    module AgreementHelper
+      def create_agreement_response(agreement:)
```
antonyoneill, line 4: I'm not sure this is a great name for this function - I get that it does "create a response for an agreement" but it reads like it "returns a response for a created agreement".

We could call it `map_agreement_to_response` or `get_agreement_response` or something. There's an element of mapping a domain model to a presentation representation of that object.

Happy to discuss

lib/hackney/income/agreement_helper.rb
```diff
@@ -0,0 +1,28 @@
...
+          startingBalance: agreement.starting_balance,
+          amount: agreement.amount,
+          startDate: agreement.start_date,
+          frequency: agreement.frequency,
+          currentState: agreement.current_state,
+          history: agreement_state_history(agreement.agreement_states)
+        }
+      end
+
+      def agreement_state_history(agreement_states)
```
antonyoneill, line 18: Similar here..

I wonder if the module should be called `AgreementResponseHelper` or something

lib/hackney/income/create_agreement.rb
```diff
@@ -0,0 +1,34 @@
+module Hackney
+  module Income
+    class CreateAgreement
+      include AgreementHelper
+
+      def execute(new_agreement_params:)
```
antonyoneill, line 6: What's the purpose of the `:`?

lib/hackney/income/create_agreement.rb
```diff
@@ -0,0 +1,34 @@
...
+          live_agreements.each do |agreement|
+            Hackney::Income::Models::AgreementState.create!(agreement_id: agreement.id, agreement_state: 'cancelled')
+          end
+        end
+
+        new_agreement = Hackney::Income::Models::Agreement.create!(agreement_params)
+        Hackney::Income::Models::AgreementState.create!(agreement_id: new_agreement.id, agreement_state: agreement_params[:current_state])
+
+        create_agreement_response(agreement: new_agreement)
+      end
```
antonyoneill, line 31: I feel like this usecase is doing too much.

Would it be a good idea to refactor out a `cancel_live_agreements` usecase?

lib/hackney/income/create_agreement.rb
antonyoneill, line 6: ~Should this usecase return a domain model (i.e. a `Agreement`) instead of a secondary representation of an Agreement?

Currently this use case is coupled with the response from the API, which in itself is not necessarily a bad thing (there are no negative implications as it stands) - but from a Clean Architecture point of view I think it would be good to decouple here~

I elaborated in the controller file

app/controllers/agreements_controller.rb
```diff
@@ -7,6 +7,20 @@ def index
...
+  def create
+    agreement_params = {
+      tenancy_ref: params.fetch(:tenancy_ref),
+      agreement_type: params.fetch(:agreement_type),
+      amount: params.fetch(:amount),
+      start_date: params.fetch(:start_date),
+      frequency: params.fetch(:frequency)
+    }
+
+    response = income_use_case_factory.create_agreement.execute(new_agreement_params: agreement_params)
```
antonyoneill, line 19: I think this controller action doesn't have quite the right balance of responsibility.

Currently it does the following:
1. Mapping an external REST request to an internal (maybe validated) request object
2. The internal request object is passed to a usecase to create the agreement
3. The response of the usecase is sent directly back over REST to the client

I think the usecase should not be aware of the interface that it has been invoked from, the burden of mapping the agreement to a REST response should be within this controller action, not of the usecase.

1. Mapping an external REST request to an internal (maybe validated) request object
2. The internal request object is passed to a usecase to create the agreement
3. Receive a new agreement object from the create agreement usecase
4. Map the new agreement to a REST safe response.

I appreciate the `index` action above doesn't follow this pattern - I think it'd be a nice refactor for a separate PR

lib/hackney/income/use_case_factory.rb
C-gyorfi: Previously it was a static class but we changed the implementation to use `AgreementHelper`, we had to do this because rais's `include` is called when a class is instantiated. 
Maybe this explains it a bit better: https://dev.to/adamlombard/ruby-class-methods-vs-instance-methods-4aje

antonyoneill: Ah! Interesting, great thanks :)

lib/hackney/income/agreement_helper.rb
C-gyorfi: I agree, probably `map_agreement_to_response` would be more clear!

lib/hackney/income/agreement_helper.rb
C-gyorfi: `map_ agreement_state_history` ?

lib/hackney/income/create_agreement.rb
C-gyorfi: Its a named argument, not sure if that's the actual term but
```Ruby
def method(named_argument:)
end

method(named_argument: 'stuff')
```

vs

```Ruby
def method(argument)
end

method('stuff')
```

C-gyorfi: Its another question weather we need to use named argument in this case

lib/hackney/income/create_agreement.rb
C-gyorfi: I agree it's certainly an SRP violation, also maybe refactor later when we would use it elsewhere?

app/controllers/agreements_controller.rb
C-gyorfi: Yes, these are really good points. I remember when I added the `index` action I pretty much followed the existing pattern in `CasesController` with the `view_cases` use-case:
https://github.com/LBHackney-IT/lbh-income-api/blob/master/app/controllers/cases_controller.rb#L5

lib/hackney/income/create_agreement.rb
antonyoneill: Oh right.

In other languages you'd use a `:` preceeded with a type declaration, othertimes a default value for the parameter.

In Kotlin you can name your parameters, or not.

Thanks again

app/controllers/agreements_controller.rb
antonyoneill: Yep, I thought that would be the case.

I'm happy with this being merged in today (I don't want us to spend too much time rewriting things just for the sake of being 'by the book')

I do think it's important to challenge existing patterns in scenarios that they might not be as maintainable as we should be striving to be. Let's just be aware of these boundaries and when we come to these kinds of controller/actions in the future let's make those boundaries clearer

spec/requests/agreements_spec.rb
```diff
@@ -76,38 +76,38 @@
         }
       end
 
-      let(:map_agreement_to_response) do
-        {
+      let(:created_agreement) do
+        Hackney::Income::Models::Agreement.create(
           id: 1,
```
C-gyorfi, line 101: We don't need to pass in the `id` this will be generated when we create a new Agreement

spec/requests/agreements_spec.rb
```diff
@@ -1,53 +1,135 @@
...
+      let(:agreement_type) { 'formal' }
+      let(:starting_balance) { Faker::Commerce.price(range: 10...1000) }
+      let(:amount) { Faker::Commerce.price(range: 10...100) }
+      let(:start_date) { Faker::Date.between(from: 2.days.ago, to: Date.today) }
+      let(:frequency) { 'weekly' }
+      let(:current_state) { 'active' }
+      let(:agreements_array) do
+        [
+          Hackney::Income::Models::Agreement.create(
             id: 1,
```
C-gyorfi, line 17: Same here, we don't need to define the `id`

spec/requests/agreements_spec.rb
```diff
@@ -1,53 +1,135 @@
...
+          tenancy_ref: tenancy_ref,
+          agreement_type: agreement_type,
+          amount: amount.to_s,
+          start_date: start_date.to_s,
+          frequency: frequency
+        }
+      end
+
+      let(:created_agreement) do
+        Hackney::Income::Models::Agreement.create(
```
C-gyorfi, line 100: Using the `create` method is a bit expensive because it actually hitting the DB, in this instance you could use `new` instead
```suggestion
        Hackney::Income::Models::Agreement.new(
```

At some point, we could set up [Factory bot](https://github.com/thoughtbot/factory_bot) and replace these to `create` method with `build` to stub them. Don't worry about this for now, we could come back to this when looking into speeding up the tests.

Ninamma: I tried this because it does make sense, but found that it broke the second test for that endpoint - the one checking if states are mapped, because in the test, we create AgreementStates and you need an Agreement from the db I guess? 

So I've stuck to Create but yeah I guess it could be improved in the other ticket.

PR #414, opened 2020-06-23
Replace BankHolidaysRetriever with NextWorkingDayRetriever

Ninamma:
And removing logic from preview generator into NextWorkingDayRetriever

## Context

[Previously](https://github.com/LBHackney-IT/lbh-income-api/pull/378), we added BankHolidaysRetriever to fix a bug with letters not taking weekends/bank holidays into account when creating the preview.

@elena-vi made the [suggestion](https://github.com/LBHackney-IT/lbh-income-api/pull/378#discussion_r430954888) to instead have a NextWorkingDayRetriever.

Since we _might_ need to use this logic for Agreements (e.g. payment being due on a Sunday or Bank Holiday might not make sense), thought it might be good to sort this out now.

- Replace BankHolidaysRetriever with NextWorkingDayRetriever
- Move logic that actually checks if a date is valid from preview generator into NextWorkingDayRetriever

## Guidance to review

I checked it works by:
- Running `make run-all`
- Using Timecop to freeze time in NextWorkingDayRetriever
image

- Trying to send a Letter 1/2 in arrears to someone

You should hopefully see something like this:
image

Please suggest any improvements that come to mind about the code! 🙏 
## Things to check


lib/hackney/pdf/next_working_day_retriever.rb
```diff
@@ -30,6 +32,14 @@ def get_dates
       def raise_error(response)
         raise UnsuccessfulRetrievalError, "Retrieval Failed: #{response.message} (#{response.code || response.status})"
       end
+
+      def get_next_working_day(bank_holidays)
+        possible_date = Time.now + 1.day
+        possible_date += 2.day if possible_date.saturday?
+        possible_date += 1.day if possible_date.sunday?
+        possible_date += 1.day while bank_holidays.include?(possible_date.strftime('%Y-%m-%d')) == true
```
C-gyorfi, line 40: Can we add more tests around this logic? I have a fear about this not tested a bit more "exhaustively". Is bank_holidays includes all the weekend days? If not, then I think this would fail when `possible_date` is let's say bank holiday Friday. In this case, the first two conditions would be ignored. And because of the third condition:
```Ruby
possible_date += 1.day while bank_holidays.include?
```
it would add 1 more day and return Saturday.

Ninamma: Ah, thank you for pointing it out! 🙏 

I'll add more tests to make sure that doesn't happen.

lib/hackney/pdf/next_working_day_retriever.rb
```diff
@@ -0,0 +1,48 @@
...
+        data.dig(DEFAULT_GROUP, 'events')&.pluck('date')
+      end
+
+      def raise_error(response)
+        raise UnsuccessfulRetrievalError, "Retrieval Failed: #{response.message} (#{response.code || response.status})"
+      end
+
+      def get_next_working_day(bank_holidays)
+        possible_date = Time.now + 1.day
+        return possible_date if bank_holidays.empty?
```
C-gyorfi, line 38: I don't understand this bit, why would we want to return the `possible_date ` if `bank_holidays` is empty? We'd still want to check for weekends?

C-gyorfi: Yep, this test failed when I removed line 38 for a good reason:
```Ruby
  context 'when Bank Holidays API request responds with 200 but body is empty' do
    it "returns tomorrow's date" do
      stub_request(:get, 'https://www.gov.uk/bank-holidays.json').to_return(
        status: 200,
        body: '{}'
      )
      Rails.cache.delete('Hackney/PDF/BankHolidays')

      next_working_day = described_class.new.execute
      expect(next_working_day).to eq((Time.now + 1.day).strftime('%d %B %Y'))
    end
  end
```
I think we should use `Timecop.freeze(some time)` around these tests otherwise the tests will have different results based on what day we are running them. E.g today is Friday so the expected time
`(Time.now + 1.day).saturday?` is true

lib/hackney/pdf/next_working_day_retriever.rb
```diff
@@ -0,0 +1,48 @@
...
+      end
+
+      def raise_error(response)
+        raise UnsuccessfulRetrievalError, "Retrieval Failed: #{response.message} (#{response.code || response.status})"
+      end
+
+      def get_next_working_day(bank_holidays)
+        possible_date = Time.now + 1.day
+
+        possible_date += 1.day while bank_holidays.include?(possible_date.strftime('%Y-%m-%d')) == true
```
C-gyorfi, line 39: Good morning 👋  Thanks for fixing this logic, just some more refactoring suggestion, it would be nice to cut down those if statements, maybe we could use helper methods to improve readability.

```Ruby
      def get_next_working_day
        possible_date = Time.now + 1.day
        possible_date += 1.day while bank_holiday?(possible_date) || weekend?(possible_date)

        possible_date
      end

      def bank_holiday?(date)
        bank_holidays.include?(date.strftime('%Y-%m-%d'))
      end

      def weekend?(date)
        date.saturday? || date.sunday?
      end

      def bank_holidays
        @bank_holidays ||= Rails.cache.fetch('Hackney/PDF/BankHolidays', expires_in: 1.day) do
          get_bank_holidays
        end
      end
```

Also using the helper method `bank_holidays` would be more optimal because it would fetch bank holidays only once in a loop.

lib/hackney/pdf/next_working_day_retriever.rb
```diff
@@ -0,0 +1,48 @@
...
+      DEFAULT_GROUP = 'england-and-wales'.freeze
+
+      def execute
+        bank_holidays = Rails.cache.fetch('Hackney/PDF/BankHolidays', expires_in: 1.day) do
+          get_bank_holidays
+        end
+
+        get_next_working_day(bank_holidays).strftime('%d %B %Y')
+      end
+
```
C-gyorfi, line 15: Can we make these methods `private` ?

spec/lib/hackney/pdf/next_working_day_retriever_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+      Timecop.freeze(2020, 5, 22)
+      next_working_day = described_class.new.execute
+      expect(next_working_day).to eq('26 May 2020')
+      Timecop.return
+    end
+  end
+
+  context 'when tomorrow is a bank holiday Friday and the following Monday is a working day' do
+    it "returns Monday's date" do
+      Timecop.freeze(2020, 5, 7)
```
C-gyorfi, line 65: Good job for using Timecop ⌚ 👍

PR #417, opened 2020-06-29
Service charge sync

elena-vi:
## Context

Adding a service charge sync query to sync over one case (pt 1 or several)

- new universal housing criteria in the service charge name space

## Guidance to review

Please look at this SQL, ask questions too
https://github.com/LBHackney-IT/lbh-income-api/pull/417/files#diff-07a88745c44a0716835bc47f2d434ac0R56

## Things to check


spec/lib/hackney/service_charge/universal_housing_criteria_spec.rb
```diff
@@ -0,0 +1,213 @@
...
+      create_uh_tenancy_agreement(
+        tenancy_ref: tenancy_ref,
+        current_balance: current_balance,
+        u_saff_rentacc: payment_ref,
+        house_ref: house_ref,
+        prop_ref: property_ref
+      )
+    }
+
+    it { is_expected.to be_instance_of(described_class) }
```
C-gyorfi, line 29: Just wondering what is `is_expected` referring to in this context? is it the `subject`?

lib/hackney/service_charge/universal_housing_criteria.rb
```diff
@@ -0,0 +1,123 @@
...
+
+          DECLARE @LastLetterDate SMALLDATETIME = (
+            #{build_last_letter_sql_query(column: 'action_date')}
+          )
+
+          DECLARE @DirectDebitStatus VARCHAR(60) = (
+            SELECT TOP 1
+              lookup.lu_desc
+            FROM
+              ddagacc
```
C-gyorfi, line 87: These table names :O

lib/hackney/service_charge/universal_housing_prioritisation_gateway.rb
```diff
@@ -0,0 +1,33 @@
...
+          criteria_start_time = Time.zone.now
+
+          criteria = Hackney::Income::UniversalHousingCriteria.for_lease(database, payment_ref)
+
+          logger(">> Time taken for Criteria from UH: #{Time.zone.now - criteria_start_time}ms")
+
+          { criteria: criteria }
+        end
+
+        logger("> Overall time taken loading from UH: #{Time.zone.now - overall_start_time}ms")
```
C-gyorfi, line 19: 👍  for these loggings

spec/support/stubs/stub_service_charge_criteria.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    def latest_letter_date
+      attributes_of_sql_result[:latest_letter_date]
+    end
+
+    def direct_debit_status
+      attributes_of_sql_result[:direct_debit_status].strip
+    end
+
+    private
```
C-gyorfi, line 43: Is this needed?

elena-vi: will be needed later of testing classification, could also be taken out for now

spec/lib/hackney/service_charge/universal_housing_criteria_spec.rb
elena-vi: yes, it's the subject

lib/hackney/service_charge/universal_housing_prioritisation_gateway.rb
```diff
@@ -0,0 +1,33 @@
+module Hackney
+  module ServiceCharge
+    class UniversalHousingPrioritisationGateway
+      def priorities_for_lease(payment_ref)
+        logger("> About to connect to UH for #{tenancy_ref}")
```
antonyoneill, line 5: I'm probably being dumb. Where does the `tenancy_ref` come from?

lib/hackney/service_charge/universal_housing_prioritisation_gateway.rb
```diff
@@ -0,0 +1,33 @@
...
+        overall_start_time = Time.zone.now
+
+        response = Hackney::UniversalHousing::Client.with_connection do |database|
```
antonyoneill, line 8: This is interesting.. How many client databases do we get here? 

I understand the what, I'm just struggling to understand the why..

lib/hackney/service_charge/universal_housing_prioritisation_gateway.rb
```diff
@@ -0,0 +1,33 @@
...
+
+        logger("> Overall time taken loading from UH: #{Time.zone.now - overall_start_time}ms")
+
+        response
+      end
+
+      private
+
+      def logger(message)
+        Rails.logger.tagged('UH-PrioritisationGateway') do
```
antonyoneill, line 27: Didn't know you could do this :ok_hand:

lib/hackney/tenancy/action_codes.rb
```diff
@@ -124,6 +124,10 @@ module ActionCodes
         STAGE_ONE_COMPLETE, STAGE_THREE_COMPLETE, COURT_BREACH_LETTER_SENT, VISIT_MADE,
         INFORMAL_BREACH_LETTER_SENT
       ].freeze
+
+      FOR_UH_LEASEHOLD_SQL = [
```
antonyoneill, line 128: This file is in the tenancy namespace - would it be clearer to have an action_codes file within the leasehold namespace?

spec/lib/hackney/service_charge/universal_housing_criteria_spec.rb
antonyoneill: This has always confused me about rspec :see_no_evil:

lib/hackney/service_charge/universal_housing_prioritisation_gateway.rb
elena-vi: this will be coming from the second sql query that I am writing right now which will be identifying which cases are in arrears and need to synced

lib/hackney/tenancy/action_codes.rb
elena-vi: I'd like to discuss this in our "how are we organising things" meeting on friday

lib/hackney/service_charge/universal_housing_prioritisation_gateway.rb
elena-vi: just one and the why is because dependancy injection (also full disclosure, I'm just reusing bits of the rents syncing process)

lib/hackney/service_charge/universal_housing_prioritisation_gateway.rb
antonyoneill: OK - So if this is merged in, and this code is executed then it'll blow up?

Is it tested? Can we either put something in here to prevent it blowing up, with a comment that it's coming soon, or wait for it to be implemented before merging? 

I'm keen that each commit is self contained and functional - leaving in a syntax error like this makes me think the commit was not finished

lib/hackney/service_charge/universal_housing_prioritisation_gateway.rb
antonyoneill: No worries about reuse of patterns - I'm a big fan of challenging existing patterns :grimacing:

I've just realised actually what this is doing. I'm used to seeing this syntax pattern for passing a lambda into a foreach on an array. Actually, you're passing a callback into the with_connection function. no probs.

lib/hackney/tenancy/action_codes.rb
antonyoneill: Perfect - I was hoping you'd say that

lib/hackney/service_charge/universal_housing_prioritisation_gateway.rb
elena-vi: this is tested and it's not being called anywhere right now, so 0 chance of anything blowing up 🤞

PR #418, opened 2020-06-30
Update get agreements response

C-gyorfi:
## Context
Agreements details page show information about an agreement `created_by` and `created_at`, currently the API endpoints does not take or return these.

- Add migration to add `created_by` field to agreements table
- Update agreement use-cases and request to return these fields 

## Link to Jira card
maybe?
https://hackney.atlassian.net/secure/RapidBoard.jspa?rapidView=30&projectKey=MAAP&modal=detail&selectedIssue=MAAP-173

## Things to check

### DIFF
--- a/spec/models/hackney/income/models/agreement_spec.rb
+++ b/spec/models/hackney/income/models/agreement_spec.rb
@@ -1,6 +1,8 @@
 require 'rails_helper'
 
 describe Hackney::Income::Models::Agreement, type: :model do
+  let(:user_name) { Faker::Name.name }
+
   it 'includes the fields for a formal/informal agreement' do
     agreement = described_class.new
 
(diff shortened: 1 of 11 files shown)


spec/models/hackney/income/models/agreement_spec.rb
```diff
@@ -1,6 +1,8 @@
 require 'rails_helper'
 
 describe Hackney::Income::Models::Agreement, type: :model do
+  let(:user_id) { Faker::Lorem.characters(number: 8) }
```
elena-vi, line 4: quick question: where is this user_id coming from? 
for documents we just store the persons name because we wanted to avoid managing users as google auth does that for us. was there a reason why we chose to store user id instead of just their name here?

C-gyorfi: Apparently there is an ID for `users`:
image

Initially, it was `<officer ID>` in the design, but I think its changed since that to user_name, I thought it's a good thing to keep the data anonymised where possible, but you are right, best if we keep it simple and follow the same pattern as we did with documents. I'll change this to record user names!
image

PR #419, opened 2020-07-01
 Add cancel endpoint to agreements API

C-gyorfi:
## Context

We want to allow caseworkers to cancel a live agreement

- Update API docs to add the new endpoint
- Add the `agreements/{id}/cancel` endpoint
- Refactor Agreements to check for any active state
- Update `create agreement` implementation so that we cancel all active agreements when creating a new agreement

## Guidance to review

Please before reviewing this have a look at this smaller PR https://github.com/LBHackney-IT/lbh-income-api/pull/418 

## Link to Jira card

https://hackney.atlassian.net/secure/RapidBoard.jspa?rapidView=30&projectKey=MAAP&modal=detail&selectedIssue=MAAP-173

## Things to check

### DIFF
--- a/docs/api/v1/api.yaml
+++ b/docs/api/v1/api.yaml
@@ -1,6 +1,6 @@
 swagger: '2.0'
 info:
-  version: '1.0'
+  version: "1.1"
   title: 'LBH Income API'
   description: 'The Income API is the back-end of Hackney Income Collection Service'
 paths:        
@@ -41,6 +41,30 @@ paths:
         required: true
         type: string
         description: Agreement for tenancy
+      responses:
+        200:
+          description: successful operation
+          schema:
+            $ref: '#/definitions/agreement'
+        400:
+          description: Invalid search parameter
+        404:
+          description: Agreement not found
+      tags:
+        - agreements
+  /agreements/{agreement_id}/cancel:
+    post:
+      summary: 'Cancel an agreement with id'
+      description: Cancel agreement with specified id
+      operationId: cancelAgreement
+      produces:
+      - application/json
+      parameters:
+      - in: path
+        name: agreement_id
+        required: true
+        type: string
+        description: Id of an agreement
       responses:
         200:
           description: successful operation
--- a/lib/hackney/income/cancel_agreement.rb
+++ b/lib/hackney/income/cancel_agreement.rb
@@ -1,0 +1,15 @@
+module Hackney
+  module Income
+    class CancelAgreement
+      def execute(agreement_id:)
+        agreement = Hackney::Income::Models::Agreement.find(agreement_id)
+
+        return agreement if agreement.current_state == 'cancelled'
+
+        Hackney::Income::Models::AgreementState.create!(agreement_id: agreement_id, agreement_state: 'cancelled')
+
+        agreement
+      end
+    end
+  end
+end
--- a/lib/hackney/income/cancel_agreement.rb
+++ b/lib/hackney/income/cancel_agreement.rb
@@ -2,8 +2,9 @@ module Hackney
   module Income
     class CancelAgreement
       def execute(agreement_id:)
-        agreement = Hackney::Income::Models::Agreement.find(agreement_id)
+        agreement = Hackney::Income::Models::Agreement.find_by_id(agreement_id)
 
+        return if agreement.nil?
         return agreement unless agreement.active?
 
         Hackney::Income::Models::AgreementState.create!(agreement_id: agreement_id, agreement_state: :cancelled)
--- a/lib/hackney/income/cancel_agreement.rb
+++ b/lib/hackney/income/cancel_agreement.rb
@@ -4,9 +4,9 @@ module Hackney
       def execute(agreement_id:)
         agreement = Hackney::Income::Models::Agreement.find(agreement_id)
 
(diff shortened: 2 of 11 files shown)


spec/lib/hackney/income/cancel_agreement_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+
+describe Hackney::Income::CancelAgreement do
+  subject { described_class.new }
+
+  let(:tenancy_ref) { Faker::Number.number(digits: 2).to_s }
+  let(:agreement_type) { 'informal' }
+  let(:amount) { Faker::Commerce.price(range: 10...100) }
+  let(:start_date) { Faker::Date.between(from: 2.days.ago, to: Date.today) }
+  let(:frequency) { 'weekly' }
+  let(:created_by) { Faker::Number.number(digits: 8).to_s }
```
elena-vi, line 11: quick question: why is this a number? 
for documents we just store the persons name because we wanted to avoid managing users as google auth does that for us

Gemfile.lock
```diff
@@ -387,7 +387,7 @@ DEPENDENCIES
   wkhtmltopdf-binary
 
 RUBY VERSION
-   ruby 2.7.1p83
+   ruby 2.5.3p105
```
antonyoneill, line 390: Whoops - I missed this :facepalm: thanks

docs/api/v1/api.yaml
```diff
@@ -1,6 +1,6 @@
 swagger: '2.0'
```
antonyoneill, line 1: Do we host these docs somewhere? Would love to take a look

docs/api/v1/api.yaml
```diff
@@ -52,6 +52,30 @@ paths:
           description: Agreement not found
       tags:
         - agreements
+  /agreements/{agreement_id}/cancel:
+    post:
```
antonyoneill, line 56: I wonder if `DELETE /agreements/{agreement_id}` would have been a better endpoint for this, maybe this is the rails way..

lib/hackney/income/cancel_agreement.rb
```diff
@@ -0,0 +1,15 @@
+module Hackney
+  module Income
+    class CancelAgreement
+      def execute(agreement_id:)
+        agreement = Hackney::Income::Models::Agreement.find(agreement_id)
```
antonyoneill, line 5: What happens if there is no agreement by this agreement_id?

lib/hackney/income/create_agreement.rb
```diff
@@ -2,28 +2,29 @@ module Hackney
...
 
-        live_agreements = Hackney::Income::Models::Agreement.where(tenancy_ref: new_agreement_params[:tenancy_ref]).where(current_state: 'live')
+        active_agreements = Hackney::Income::Models::Agreement.where(tenancy_ref: tenancy_ref).select(&:active?)
 
-        if live_agreements.any?
-          live_agreements.each do |agreement|
-            Hackney::Income::Models::AgreementState.create!(agreement_id: agreement.id, agreement_state: 'cancelled')
+        if active_agreements.any?
+          active_agreements.each do |agreement|
+            Hackney::Income::Models::AgreementState.create!(agreement_id: agreement.id, agreement_state: :cancelled)
```
antonyoneill, line 22: Can this usecase use the `CancelAgreement` usecase?

spec/lib/hackney/income/cancel_agreement_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+      created_by: created_by
+    }
+  end
+
+  let!(:agreement) { Hackney::Income::Models::Agreement.create!(agreement_params) }
+
+  before do
+    Hackney::Income::Models::AgreementState.create(agreement_id: agreement.id, agreement_state: 'live')
+  end
+
```
antonyoneill, line 29: Would be good to have a test to what happens if you try to cancel an agreement that doesn't exist, also what about a breached agreement?

docs/api/v1/api.yaml
C-gyorfi: Yes and no, its hosted with the income-api, but the API accepts requests from the Front-end application, so you won't be able to browse it in prod(this is not ideal but I wasn't aware of this when I added the docs). 

Otherwise when you are running the income-api or all the application using `make run-all` locally you can browse it just by visiting your localhost

docs/api/v1/api.yaml
C-gyorfi: I was not sure what is the convention, so I followed the pattern we used on another rails API.

spec/lib/hackney/income/cancel_agreement_spec.rb
C-gyorfi: I think we want to be able to cancel breached agreements, I've added some triangulation to make this clear in 883c819

lib/hackney/income/create_agreement.rb
C-gyorfi: Resolved in 6c07be1

lib/hackney/income/cancel_agreement.rb
C-gyorfi: Thanks for pointing this out, resolved in: ea2076f

PR #420, opened 2020-07-01
sql to identify what cases need to be synced

elena-vi:
## Context

Adding a service charge sync query to sync over one case (pt 2 of several)

- Added sql query that will identify what cases need to be queued up and processed by the _new_ service charge universal housing criteria (implemented in https://github.com/LBHackney-IT/lbh-income-api/pull/417)

## Guidance to review

- The main sql that identifies the cases is here: `lib/hackney/service_charge/universal_housing_lease_gateway.rb` 

## Things to check

### DIFF
--- a/spec/lib/hackney/service_charge/universal_housing_lease_gateway_spec.rb
+++ b/spec/lib/hackney/service_charge/universal_housing_lease_gateway_spec.rb
@@ -27,7 +27,7 @@ describe Hackney::ServiceCharge::UniversalHousingLeaseGateway, universal: true d
         create_uh_tenancy_agreement(tenancy_ref: '000003/01', current_balance: 50.00)
       end
 
-      it 'returns only the master account tenancy' do
+      it 'returns only the one in lease rent group' do
         expect(subject).to eq(['000001/01'])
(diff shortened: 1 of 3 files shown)


spec/lib/hackney/service_charge/universal_housing_lease_gateway_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+    end
+
+    context 'when there are three tenancies in arrears, but only one is in the leasehold rent group' do
+      before do
+        create_uh_tenancy_agreement(tenancy_ref: '000001/01', current_balance: 50.00, rentgrp_ref: lease_rent_group)
+        create_uh_tenancy_agreement(tenancy_ref: '000002/01', current_balance: 50.00)
+        create_uh_tenancy_agreement(tenancy_ref: '000003/01', current_balance: 50.00)
+      end
+
+      it 'returns only the master account tenancy' do
```
C-gyorfi, line 30: I think you meant to 
```suggestion
      it 'returns only the one in lease rent group' do
```

PR #421, opened 2020-07-06
move leasehold things into new namsespace

elena-vi:
## Context

Based on the discussion we all had about how code should be structured we decided to move all the leasehold things to be in the leasehold namespace

- rename `service_charge` to `leasehold` 

## Things to check


app/controllers/application_controller.rb
```diff
@@ -14,7 +14,7 @@ def pdf_use_case_factory
   end
 
   def service_charge_use_case_factory
```
Ninamma, line 16: just a question really, is it because of the smol slices that we can't change this bit to 'leasehold' just yet?

lib/tasks/income/sync.rake
```diff
@@ -1,20 +1,38 @@
...
+      desc 'enqueues workers for income_use_case_factory.schedule_sync_cases.execute'
+      task :enqueue do
+        p 'pooo?'
+        use_case_factory = Hackney::Income::UseCaseFactory.new
+        use_case_factory.schedule_sync_cases.execute
+      end
+
+      desc 'manually runs the sync'
+      task :manual_sync do
+        p 'yes?'
```
Ninamma, line 14: not sure if you're leaving the printouts accdientally or because you'll need it later - pointing out just in case

elena-vi: accidentally, good catch

PR #426, opened 2020-07-14
Expose tenancy_ref in classifier log

antonyoneill:
## Context

To aid in debugging classifier rules - at the moment the log output is not useful in identifying a tenancy.

Just a small one

* Emit a tenancy ref rather than a set of object hashes.

## Link to Jira card

https://hackney.atlassian.net/secure/RapidBoard.jspa?rapidView=30&projectKey=MAAP&modal=detail&selectedIssue=MAAP-202

## Things to check

### DIFF
--- a/lib/hackney/income/universal_housing_criteria.rb
+++ b/lib/hackney/income/universal_housing_criteria.rb
@@ -1,6 +1,8 @@
 module Hackney
   module Income
     class UniversalHousingCriteria
+      attr_reader :tenancy_ref
+
       def self.for_tenancy(universal_housing_client, tenancy_ref)
         attributes = universal_housing_client[build_sql, tenancy_ref].first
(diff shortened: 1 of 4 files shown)


lib/hackney/income/universal_housing_criteria.rb
```diff
@@ -264,7 +267,7 @@ def self.beginning_of_week
 
       private
 
-      attr_reader :tenancy_ref, :attributes
+      attr_reader :attributes
```
C-gyorfi, line 270: A bit confused, why do we have 2 `attr_reader` ? 🤔

antonyoneill: `:attributes` was private before hand, I didn't want to change that - idk why it is private though..

PR #430, opened 2020-07-14
Check if active court warrant before sending nosp

antonyoneill:
## Context

The old classification engine had very loose rules. As a result we can trigger several actions for a given case, even though it should not.

Tighten the rules for sending NOSPs. In the this PR I have prevented a NOSP being sent if one of 6 court outcomes occurred within the last 6 years.

Technically some of them would have custom dates, but they are edge cases, and the dates are currently not stored in a normalised form in UH. Once we start modelling court outcomes and strike out dates then we can revisit.

## Guidance to review

Check if the send_nosp ruleset seems sensible to you, given the knowledge of the escalation process.

Incidentally, this resolves 19/21 deviations in this scenario. 1 of them is a potential out-of-date sync, and the other is a missing court outcome from the model "CDG". If you know what CDG means, please let me know.

## Link to Jira card

https://hackney.atlassian.net/secure/RapidBoard.jspa?rapidView=30&projectKey=MAAP&modal=detail&selectedIssue=MAAP-202

## Things to check

### DIFF
--- a/lib/hackney/tenancy/court_outcome_codes.rb
+++ b/lib/hackney/tenancy/court_outcome_codes.rb
@@ -3,6 +3,12 @@ module Hackney
     module CourtOutcomeCodes
       OUTRIGHT_POSSESSION_WITH_DATE = 'OPD'.freeze
       OUTRIGHT_POSSESSION_FORTHWITH = 'OPF'.freeze
+
+      ADJOURNED_GENERALLY = 'AGE'.freeze
+      ADJOURNED_ON_TERMS = 'ADT'.freeze
+      ADJOURNED_ON_TERMS_SECONDARY = 'AOT'.freeze
+
+      SUSPENDED_POSSESSION = 'SUP'.freeze
     end
   end
 end
--- a/lib/hackney/income/tenancy_classification/v2/helpers.rb
+++ b/lib/hackney/income/tenancy_classification/v2/helpers.rb
@@ -11,6 +11,23 @@ module Hackney
             @criteria.eviction_date.present?
           end
 
+          def court_warrant_active?
+            return false if @criteria.court_outcome.blank?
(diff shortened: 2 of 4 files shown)


lib/hackney/tenancy/court_outcome_codes.rb
```diff
@@ -3,6 +3,12 @@ module Tenancy
     module CourtOutcomeCodes
       OUTRIGHT_POSSESSION_WITH_DATE = 'OPD'.freeze
       OUTRIGHT_POSSESSION_FORTHWITH = 'OPF'.freeze
+
+      ADJOURNED_GENERALLY = 'AGE'.freeze
+      ADJOURNED_ON_TERMS = 'ADT'.freeze
+      ADJOURNED_ON_TERMS_SECONDARY = 'AOT'.freeze
+
+      SUSPENDED_POSSESSION = 'SUP'.freeze
```
C-gyorfi, line 11: Silly question, where did you get these codes and definitions from?

antonyoneill: There are no silly questions..

I ended up cross referencing what is defined here:

https://github.com/LBHackney-IT/lbh-income-api/blob/1bc375e9c394002845f565dce3e7ab236bbba72e/lib/hackney/tenancy/action_codes.rb#L55-L57

https://github.com/LBHackney-IT/LBH-IncomeCollection/blob/575cbd4b04f7e8d54d1b0e8a76039e608a92818a/lib/hackney/income/action_diary_entry_codes.rb#L62-L63

It does bother me that we have them defined in multiple places.

C-gyorfi: Yeah, would be nice to have one source of truth

PR #436, opened 2020-07-15
HIGH PRIORITY: Fix balance on income letters

elena-vi:
## Context

When we send our arrears letters we need to take the balance from the previous period end.

Letter need to cover the previous week’s rent (not including this week’s rent).

- added a new collectable arrears column to case priorities table
- refactored the letters to use the synced calculated `collectable_arrears` value 

## Guidance to review

I decided to import the stored tenancy gateway in [lib/hackney/pdf/income_preview.rb](https://github.com/LBHackney-IT/lbh-income-api/blob/a012469f8ae17a6569ab4661c090e4155a3a47b7/lib/hackney/pdf/income_preview.rb) instead of modifying the SQL in [lib/hackney/income/universal_housing_income_gateway.rb](https://github.com/LBHackney-IT/lbh-income-api/blob/a012469f8ae17a6569ab4661c090e4155a3a47b7/lib/hackney/income/universal_housing_income_gateway.rb). I thought that it would be worth to save the magic calculated `collectable_arrears` rather than fetch it once just for the letter (like what we're doing to find the address in [lib/hackney/income/universal_housing_income_gateway.rb](https://github.com/LBHackney-IT/lbh-income-api/blob/a012469f8ae17a6569ab4661c090e4155a3a47b7/lib/hackney/income/universal_housing_income_gateway.rb))

## Link to Jira card

https://hackney.atlassian.net/browse/MAAP-326

## Things to check


spec/lib/hackney/income/universal_housing_criteria_spec.rb
```diff
@@ -197,6 +197,44 @@
...
+        it { is_expected.to eq(current_balance) }
+      end
+
+      context 'when rent was issued on monday and a payment was made on tuesday' do
+        before do
+          create_uh_transaction(tenancy_ref: tenancy_ref, type: 'RNT',  amount: rent, date: Date.today.beginning_of_week)
+          create_uh_transaction(tenancy_ref: tenancy_ref, type: 'RPY',  amount: payment, date: Date.today.beginning_of_week + 1.day)
+        end
+
+        it { is_expected.to eq(current_balance - (rent + payment)) }
```
antonyoneill, line 216: I always question whether there should be logic in assertions like this vs using a known `:payment` and `:rent` value and an expected current balance

PR #439, opened 2020-07-17
Add create court case use-case

C-gyorfi:
## Context
We want to be able to create a court case to be able to create a formal agreement.

- Add a use-case for creating court case

## Link to Jira card
https://hackney.atlassian.net/browse/MAAP-332
## Things to check


antonyoneill:
:heart_eyes: such a small PR :raised_hands:

PR #443, opened 2020-07-20
Add GET court cases API endpoint

C-gyorfi:
## Context
We want to be able to view existing court cases so we can view the court case of a formal agreement

- Add get court cases API endpoint

## Link to Jira card

https://hackney.atlassian.net/browse/MAAP-332

## Things to check

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ serve:
 
 .PHONY: shell
 shell:
-	docker-compose run --rm -e RAILS_ENV=test app /bin/bash
+	docker-compose run --rm app /bin/bash
 
 .PHONY: test-db-destroy
 test-db-destroy:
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -21,7 +21,7 @@ serve:
 
 .PHONY: shell
 shell:
-	docker-compose run --rm -e RAILS_ENV=test app /bin/bash
+	docker-compose run --rm app /bin/bash
```
C-gyorfi, line 24: This was an accident, I had to remove the `RAILS_ENV=test` otherwise the migration trips over when I start the shell. I leave it here because I don't think we need to run the shell in test env.

elena-vi: yea, we don't need to run it in test.. (I think I might have left that in by accident)

PR #454, opened 2020-07-30
Add use-case to detect breaches in informal agreements

C-gyorfi:
## Context

This is the first slice of tackling this problem and is not covering all the edge cases, simply just updates the agreement status to breached.
Next things to implement:
- Reset the agreement status to `live` when an agreement is no longer breached
- Add some info on why the breach is detected, e.g the UI design expects to show expected balance and actual balance and date of the last check, we should record these somewhere
- Add a corn job to run the process daily

- Add a use-case that can take an agreement and determine whenever its breached or not, based on its first instalment date(`start_date`), frequency of payment, the days of tolerance for completing the payment. 

## Guidance to review
When we discussed detecting breaches previously, there was an expectation to check for breaches according to the payment cycle. Why is this hard to achieve? 
- The payment cycle can start any date
- We have multiple frequencies
- The status can change from one day to another(e.g payment being made on a day, the agreement is no longer in breach the next day)
It is better if we run batch processing nighly on all live agreements so we can see whether on that day it's breached or not.

## Link to Jira card

## Things to check

### DIFF
--- a/spec/factories/agreement.rb
+++ b/spec/factories/agreement.rb
@@ -1,0 +1,32 @@
+FactoryBot.define do
+  factory :agreement, class: Hackney::Income::Models::Agreement do
+    tenancy_ref { Faker::Lorem.characters(number: 5) }
+    agreement_type { :informal }
+    notes { Faker::ChuckNorris.fact }
+    created_by { Faker::Name.name }
+    frequency { [:weekly, :monthly, :fortnightly, '4 weekly'].sample }
+    start_date { Faker::Date.between(from: 2.days.ago, to: Date.today) }
+    amount { Faker::Commerce.price(range: 10...100) }
+  end
+
+  factory :agreement_state, class: Hackney::Income::Models::AgreementState do
+    agreement
+    agreement_state { :live }
+
+    trait :live do
+      agreement_state { :live }
+    end
+
+    trait :breached do
+      agreement_state { :breached }
+    end
+
+    trait :cancelled do
+      agreement_state { :cancelled }
+    end
+
+    trait :completed do
+      agreement_state { :completed }
+    end
+  end
+end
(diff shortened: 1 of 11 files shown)


spec/lib/hackney/income/detect_breach_spec.rb
```diff
@@ -0,0 +1,241 @@
...
+          expected: 2
+        },
+        {
+          start_date: Time.zone.local(2019, 2, 1),
+          current_date: Time.zone.local(2019, 2, 1) + 29.days,
+          expected: 1
+        }
+      ].each do |example|
+        Timecop.freeze(example[:current_date]) do
+          expect(subject.send(:full_months_since, example[:start_date])).to eq(example[:expected])
```
Ninamma, line 207: this is cool, i've never properly seen `subject.send` before - I think. 😅 
Is this used to just target/test the method `full_months_since` in the subject?

Ninamma: ohhh i see - that's good to know, thanks!

spec/factories/agreement.rb
```diff
@@ -0,0 +1,32 @@
...
+    agreement_type { :informal }
+    notes { Faker::ChuckNorris.fact }
+    created_by { Faker::Name.name }
+    frequency { [:weekly, :monthly, :fortnightly, '4 weekly'].sample }
+    start_date { Faker::Date.between(from: 2.days.ago, to: Date.today) }
+    amount { Faker::Commerce.price(range: 10...100) }
+  end
+
+  factory :agreement_state, class: Hackney::Income::Models::AgreementState do
+    agreement
```
Ninamma, line 13: all looks good! 

just for my understanding, could you explain a bit about putting `agreement` here, instead of maybe just `agreement_id` - is this to do with the relationship between agreements and agreement states?

Or is it to allow you to give it either an `agreement` (like here)
image

 or just an `agreement_id` like a lot of places e.g. here
image


I tried to figure out how to link to those lines but I couldn't 😅

C-gyorfi: Yes, it comes from the association, you can check the agreement on a state just by calling agreement:

```Ruby
state.agreement
```

likewise, you can create the state by passing an associated agreement and probably its the simplest way to do it. Thanks for flagging the inconsistency

PR #457, opened 2020-07-31
Update agreement state when the state has changed since last check

C-gyorfi:
## Context

This is the second slice of tackling this problem, see the first slice in https://github.com/LBHackney-IT/lbh-income-api/pull/454 

Next things to implement:
- Add a cron job to run the process daily
- updating API endpoint with the full agreement_status response so we can show the state on the front-end
- Anything else I forgot?

- Add `current_balance`, `expected_balance` and `description` fields to `agreement_states` table, we want to show this data on the UI
- Add migration to fix the `scale` wherever we store money in `decimal` in DB, to avoid losing the decimal values
- Reset the agreement status to `live` when an agreement is no longer breached
- Update last checked date when the state has not changed since the previous check
- Add new state if the state has changed since the last check
- Complete agreement when the account is no longer in arrears

## Guidance to review
This class is growing big and I'd like to make sure it's clear and maintainable, what do you think? There is quite a bit of logic in there, shall we split this into multiple use-cases instead?

## Link to Jira card


lib/hackney/income/detect_breach.rb
```diff
@@ -57,6 +70,35 @@ def payment_cycles_since(date, days)
       def prevent_negative(number)
         [number, 0].max
       end
+
+      def date_of_first_check(agreement)
+        agreement.start_date + @tolerance_days.days
+      end
+
+      def update_last_checked_date(agreement)
+        Hackney::Income::Models::AgreementState.where(agreement_id: agreement.id).last&.touch
```
Ninamma, line 79: ooooooh - another new thing I'm learning

db/migrate/20200730154821_add_balance_fields_to_agreement_states.rb
```diff
@@ -0,0 +1,7 @@
+class AddBalanceFieldsToAgreementStates < ActiveRecord::Migration[5.2]
+  def change
+    add_column :agreement_states, :current_balance, :decimal, precision: 10, scale: 2
```
elena-vi, line 3: Just a quick question: 
why are we storing the `current_balance` on here? I'm a little bit confused why we need it here when an agreement_state is related to agreements and your can get the balance from the case/action?

C-gyorfi: You may be right, my concern is the UI should show the status with current balance and expected balance, however, if the current balance has changed since the last check, that could cause bugs such as the current balance is below the expected balance but the state is still breached because the last check date was earlier than the time the payment has gone through:

image

C-gyorfi: I think the name `current balance` does not make sense, it should be balanced when checked or something like that

C-gyorfi: Renamed the field to `checked_balance` as discussed and force pushed to 95737f6

PR #460, opened 2020-08-03
Update court case table and model

Ninamma:
## Context

We want to allow officers in MA to be able to:
- Add a court case with _just_ the court date for a case/tenancy, before the actual court case takes place
- Add an outcome (and other details) for a court case that has taken place

We also want to allow for syncing of court outcomes in UH
- Some outcomes may not have a court date
- Some existing outcomes in UH won't map to our new list of court outcomes

- Updated documentation for court case end point:
  - Added request body guidance
  - Changed structure of court cases, removed `createdBy` and added `courtCaseHistory` to let us record the history of a court case, for example:
```
{
  "id": 12,
  "tenancyRef": "1",
  "courtDate": "01/08/2020",
  "courtOutcome": "Stay of execution",
  "balanceOnCourtOutcomeDate": 1000,
  "strikeOutDate": "01/08/2026",
  "courtCaseHistory": [
       {
      "description": "court outcome added",
      "date": "04/08/2020",
      "createdBy": "Isabelle"
     },
    {
      "description": "court date added",
      "date": "04/07/2020",
      "createdBy": "Tom Nook"
    }
  ]
}
```
- Updated court cases table:
  - Changed column names to match design: `court_decision_date` -> `court_date`, `balance_at_outcome_date` -> `balance_on_court_outcome_date`
  - Allow `strike_out_date` to be null - since outcomes won't always have one (e.g. adjourned ones)
  - Remove `created_by` column, since we will record these in a court case history table (mainly since one officer might create a court case with a date and another officer will add the outcome after the court case has taken place)

- Updated CourtCase model:
   - Only validate the presence of `tenancy_ref` - since other attributes (properties? fields? sorry, not 100% sure what the correct term is) might not exist
   - `courtDecisionDate` -> `courtDate`
   - `balanceAtOutcomeDate` -> `balanceOnCourtOutcomeDate`

To do:
- Add column in court cases table for `counter claim for disrepair` - in Jenny's designs, came from discussions with Ian
- Add court case history table
- Update use case and frontend accordingly
## Guidance to review

My brain feels pretty fried right now 😅 so please let me know if anything seems silly.

## Link to Jira card

https://hackney.atlassian.net/browse/MAAP-349?atlOrigin=eyJpIjoiMzZhYmNkNmY1YzJhNDUzMGE0MGYwN2NmYmZjNGQ4NzYiLCJwIjoiaiJ9

https://hackney.atlassian.net/browse/MAAP-224?focusedCommentId=15817

## Things to check

### DIFF
--- a/db/migrate/20200804143043_update_nullable_court_case_columns.rb
+++ b/db/migrate/20200804143043_update_nullable_court_case_columns.rb
@@ -1,0 +1,6 @@
+class UpdateNullableCourtCaseColumns < ActiveRecord::Migration[5.2]
+  def change
+    change_column_null :court_cases, :court_date, true
+    remove_column :court_cases, :created_by, :string
+  end
+end
--- a/spec/models/hackney/income/models/court_cases_spec.rb
+++ b/spec/models/hackney/income/models/court_cases_spec.rb
@@ -5,16 +5,16 @@ describe Hackney::Income::Models::CourtCase, type: :model do
     court_details = described_class.new
     expect(court_details.attributes).to include(
       'tenancy_ref',
-      'court_decision_date',
+      'date_of_court_decision',
       'court_outcome',
-      'balance_at_outcome_date'
+      'balance_on_court_outcome_date'
     )
   end
 
   it { is_expected.to validate_presence_of(:tenancy_ref) }
-  it { is_expected.to validate_presence_of(:court_decision_date) }
+  it { is_expected.to validate_presence_of(:date_of_court_decision) }
   it { is_expected.to validate_presence_of(:court_outcome) }
-  it { is_expected.to validate_presence_of(:balance_at_outcome_date) }
+  it { is_expected.to validate_presence_of(:balance_on_court_outcome_date) }
   it { is_expected.to validate_presence_of(:strike_out_date) }
   it { is_expected.to validate_presence_of(:created_by) }
   it { is_expected.to have_many(:agreements) }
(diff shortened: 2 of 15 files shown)


db/migrate/20200804143043_update_nullable_court_case_columns.rb
```diff
@@ -0,0 +1,6 @@
+class UpdateNullableCourtCaseColumns < ActiveRecord::Migration[5.2]
+  def change
+    change_column_null :court_cases, :court_date, true
```
C-gyorfi, line 3: Didn't we do the opposite in the other migration file?
https://github.com/LBHackney-IT/lbh-income-api/blob/fc9f8adc9924d7a6c22e1c1d05e06c5280c30206/db/migrate/20200803165658_update_court_case_columns.rb#L11

spec/models/hackney/income/models/court_cases_spec.rb
```diff
@@ -41,4 +36,26 @@
     expect(described_class.first.agreements.first).to be_a Hackney::Income::Models::Agreement
     expect(Hackney::Income::Models::Agreement.first.court_case).to eq(court_case)
   end
+
+  context 'when there is only a court date (e.g. adding a court date before the case takes place)' do
```
C-gyorfi, line 40: These tests feel a bit redundant since they are testing the absence of validation

db/migrate/20200804143043_update_nullable_court_case_columns.rb
Ninamma: yup i think this was because i initially thought court cases would need a court date, but after we discussed that we would sync court cases from UH that might not have a court date, i had to change it back.

Maybe i should've done this differently/more cleverly and just had one migration 😅

Ninamma:
> Makes sense I guess, I'm slightly concerned about the history, do we need to track the history of every single field? Although, I'm not sure what would be a better way to do it 🤔

Hmm not sure I fully understand what you mean by this 🤔 Maybe we can have a call in the afternoon?

spec/models/hackney/income/models/court_cases_spec.rb
Ninamma: yeah that's probably true - i think i was trying to be "good" and TDD my way to removing the validation 😅 also wanted to be explicit about why we allow almost everything to be null

PR #461, opened 2020-08-04
Add cron job to update agreement states nightly

C-gyorfi:
## Context

Another slice of tackling this problem, see the previous slice in https://github.com/LBHackney-IT/lbh-income-api/pull/457

- Rename DetectBreach usecase to UpdateAgreement state as the new class name describes the responsibility of the class a bit better
- Add use-case to run update agreement state use-case on all active agreements
- Add a cron job to run the process daily at 4.00 am
- Update docs and add rake task

## Guidance to review
I don't have the confidence of this is tested very well, I spend some time understanding how do we test cron jobs but still a bit confused, is there a better way to test this? Potentially with an end-to-end test?

## Link to Jira card


lib/hackney/income/update_all_agreement_state.rb
```diff
@@ -0,0 +1,22 @@
...
+        all_active_agreements =
+          Hackney::Income::Models::Agreement.where(current_state: :live)
+                                            .or(Hackney::Income::Models::Agreement.where(current_state: :breached))
+
+        all_active_agreements.each do |agreement|
+          @update_agreement_state.execute(agreement: agreement)
+        rescue StandardError => e
+          puts "[#{Time.now}] Failed to update agreement state " \
+               "for agreement: #{agreement.id} on tenancy: #{agreement.tenancy_ref}, with error: #{e.inspect}."
+        end
```
elena-vi, line 18: it would be great to raise a sentry error 
```ruby
Raven.capture_exception(exception)
```

lib/hackney/income/update_all_agreement_state.rb
```diff
@@ -0,0 +1,22 @@
...
+    class UpdateAllAgreementState
+      def initialize(update_agreement_state:)
+        @update_agreement_state = update_agreement_state
+      end
+
+      def execute
+        all_active_agreements =
+          Hackney::Income::Models::Agreement.where(current_state: :live)
+                                            .or(Hackney::Income::Models::Agreement.where(current_state: :breached))
+
```
elena-vi, line 12: Could possibly run this query instead of line 10 + 11

```
Hackney::Income::Models::Agreement.where('current_state =? OR current_state =?', 'live', 'breached')
```

PR #463, opened 2020-08-07
Refactor SyncCasePriority 

C-gyorfi:
## Context
We want to use MAA agreements rather than the agreements from UH whenever we running classification, currently, the classifier called in the sync process, in `stored_tenancies_gateway` whenever a new `case_priority` is saved in the DB.

We want to refactor these responsibilities so we can update that agreement states in MAA before running the classifier and then saving a new worktray item(aka case priority) 

-  Rename stored_tenancies_gateway to  stored_worktray_item_gateway

## Link to Jira card

## Things to check

### DIFF
--- a/lib/hackney/income/sync_case_priority.rb
+++ b/lib/hackney/income/sync_case_priority.rb
@@ -1,15 +1,15 @@
 module Hackney
   module Income
     class SyncCasePriority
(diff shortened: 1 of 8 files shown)


antonyoneill:
Looks good - this is a very nice thin slice - can we get this merged on it's own?

PR #465, opened 2020-08-10
Refactor update agreement usecase

C-gyorfi:
## Context
We want to use MAA agreements rather than the agreements from UH whenever we running classification, currently, the classifier called in the sync process, in `stored_tenancies_gateway` whenever a new `case_priority` is saved in the DB.

We want to refactor these responsibilities so we can update that agreement states in MAA before running the classifier and then saving a new worktray item(aka case priority) 

This is the third slice of this refactoring, second slice: https://github.com/LBHackney-IT/lbh-income-api/pull/464

-  Refactor `update agreement` state usecase
- Remove `update all agreement states` worker
- Update agreements in `SyncCasePriority`

## Link to Jira card

## Things to check


schedule.yml
```diff
@@ -15,10 +15,6 @@
   leasehold_sync:
     class: LeaseholdSync
     cron: '0 0 3 * * *' # 3am everyday
-  
-  update_all_agreement_state:
```
Ninamma, line 19: just to help my understanding 😅 so now instead of a separate sync job, we put updating agreements states in the (tenancy? leasehold? both?) sync that's above - so that it runs before the classifier?

C-gyorfi: Yes, exactly, I realised later that we need to run this before the classifier 😅

lib/hackney/income/sync_case_priority.rb
```diff
@@ -1,17 +1,26 @@
...
     class SyncCasePriority
+      WorktrayItemModel = Hackney::Income::Models::CasePriority
+      DocumentModel = Hackney::Cloud::Document
+      AgreementModel = Hackney::Income::Models::Agreement
+
       def initialize(prioritisation_gateway:, stored_worktray_item_gateway:, automate_sending_letters:)
         @automate_sending_letters = automate_sending_letters
         @prioritisation_gateway = prioritisation_gateway
         @stored_worktray_item_gateway = stored_worktray_item_gateway
+        @update_agreement_state = Hackney::Income::UseCaseFactory.new.update_agreement_state
```
elena-vi, line 12: could we pass in an instance of `update_agreement_state` use case here like, just like how we pass in `automate_sending_letters`

PR #466, opened 2020-08-10
Fix create court case endpoint

Ninamma:
## Context

When an officer adding an `Adjourned` court outcome, they will be asked for two extra things:
- If the outcome has `terms`
- If there is a `disrepair counter claim`

- Added `terms` and `disrepair counter claim` coulmns to `court_case` table
- Added these fields (parameters?) to the CourtCase model, where the outcome is adjourned

- Moved some of the existing UH Court Outcome codes, so they're all in one file (where as currently they're duplicated)
- Added a similar file for Updated Court Outcome codes 
## Guidance to review

Would love a look at the way I've like added the validation in the CourtCase model for `adjourned` cases?

## Things to check


spec/factories/court_case.rb
```diff
@@ -2,7 +2,7 @@
   factory :court_case, class: Hackney::Income::Models::CourtCase do
     tenancy_ref { "#{Faker::Number.number(digits: 6)}/#{Faker::Number.number(digits: 2)}" }
     court_date { Faker::Date.between(from: 10.days.ago, to: 2.days.ago) }
-    court_outcome { Faker::ChuckNorris.fact }
```
antonyoneill, line 5: Thank you.

lib/hackney/tenancy/action_codes.rb
```diff
@@ -52,14 +52,10 @@ module ActionCodes
...
-      ADJOURNED_ON_TERMS_COURT_OUTCOME = 'ADT'.freeze
-      POSTPONED_POSSESSIOON_COURT_OUTCOME = 'PPO'.freeze
-      SUSPENDED_POSSESSION_COURT_OUTCOME = 'SPO'.freeze
-
       PAUSED_MISSING_DATA = 'RMD'.freeze
 
-      ADJOURNED_GENERALLY = 'ADG'.freeze
-      ADJOURNED_ON_TERMS = 'ADT'.freeze
+      ADJOURNED_GENERALLY = Hackney::Tenancy::CourtOutcomeCodes::ADJOURNED_GENERALLY
+      ADJOURNED_ON_TERMS = Hackney::Tenancy::CourtOutcomeCodes::ADJOURNED_ON_TERMS
```
elena-vi, line 58: this makes me happy

PR #467, opened 2020-08-10
Refactor classification helpers and add helpers for ma agreements

C-gyorfi:
## Context
We want to use MAA agreements rather than the agreements from UH whenever we running classification

- Move agreement rules to `UHAgreementHelpers` from `TenancyClassification::V2::Helper`
- Add agreements helpers for MA agreements `MAAgreementHelpers`

## Link to Jira card

## Things to check


lib/hackney/income/tenancy_classification/v2/helpers/helpers_base.rb
```diff
@@ -0,0 +1,13 @@
+module Hackney
+  module Income
+    module TenancyClassification
+      module V2
+        module Helpers
+          module HelpersBase
+            def should_prevent_action?; end
```
antonyoneill, line 7: I don't think helpers like `informal_breached_agreement` should be checking if `should_prevent_action`. The answer to those questions has nothing to do with if a case is paused etc.

elena-vi: I'm confused about the purpose of this helper, but everything else looks good

C-gyorfi: Removed the coupled `should_prevent_action?` from helper methods and removed the helper base class in 592559d

PR #469, opened 2020-08-11
Allow updating court case

Ninamma:
## Context

For an officer, adding a court case is a 2 process step in the design, they:

1.  Add the court date (most likely before the court case takes place)
2. Add the outcome once the court case had taken place

I'm sorry it's so big 😭 - definitely could and should have sliced it into use case then endpoint... 😞 
- Adds an `update_court_case` use case that takes outcome information and updates an existing court case
- Adds a PATCH endpoint for updating an existing court case

To do:
- Update documentation for the endpoint (will do as I wait for this PR to be reviewed)

## Guidance to review

- In the controller, I separated required and optional parameters
  - Is that unecessary or is having required params making things worse, since we will also check for required fields in the frontend
- I made the `update_court_case` use case return `nil` if the `court case id`, `court date` and `tenancy ref` doesn't match an those of an existing court case
   - Same question as above really, is this unecessary or does it make sense?
   - `court case id` makes sense
   - But as I type this, I am unsure about the others, since in the designs, when adding the court outcomes you don't even enter those
  - I guess I was trying to be extra safe but maybe it's _too much_
```
       return if court_case.nil?
       return if court_case.court_date != court_date
       return if court_case.tenancy_ref != tenancy_ref
```
## Link to Jira card

https://hackney.atlassian.net/browse/MAAP-411?atlOrigin=eyJpIjoiYzU4Yjk3ZjMxOTY0NGUwOWFhM2U1NGNjZmM5N2I5OWIiLCJwIjoiaiJ9
## Things to check

### DIFF
--- a/app/controllers/court_cases_controller.rb
+++ b/app/controllers/court_cases_controller.rb
@@ -1,6 +1,7 @@
 class CourtCasesController < ApplicationController
   include CourtCaseResponseHelper
-
+  REQUIRED_PARAMS = %i[id tenancy_ref court_date court_outcome balance_on_court_outcome_date].freeze
+  OPTIONAL_PARAMS = %i[strike_out_date terms disrepair_counter_claim].freeze
   def index
     requested_cases = income_use_case_factory.view_court_cases.execute(tenancy_ref: params.fetch(:tenancy_ref))
 
@@ -26,11 +27,14 @@ class CourtCasesController < ApplicationController
 
   def update
     court_case_params = {
-      id: params.require(:id),
-      tenancy_ref: params.require(:tenancy_ref),
-      court_date: params.require(:court_date),
-      court_outcome: params.require(:court_outcome),
-      balance_on_court_outcome_date: params.require(:balance_on_court_outcome_date)
+      id: update_court_case_params[:id],
+      tenancy_ref: update_court_case_params[:tenancy_ref],
+      court_date: update_court_case_params[:court_date],
+      court_outcome: update_court_case_params[:court_outcome],
+      balance_on_court_outcome_date: update_court_case_params[:balance_on_court_outcome_date],
+      strike_out_date: update_court_case_params[:strike_out_date],
+      terms: update_court_case_params[:terms],
+      disrepair_counter_claim: update_court_case_params[:disrepair_counter_claim]
     }
 
     updated_court_case = income_use_case_factory.update_court_case.execute(court_case_params: court_case_params)
@@ -38,4 +42,10 @@ class CourtCasesController < ApplicationController
 
     render json: response
   end
+
+  def update_court_case_params
+    params.require(REQUIRED_PARAMS)
+    allowed_params = params.permit(REQUIRED_PARAMS + OPTIONAL_PARAMS)
+  end
+
 end
--- a/app/controllers/court_cases_controller.rb
+++ b/app/controllers/court_cases_controller.rb
@@ -1,7 +1,8 @@
 class CourtCasesController < ApplicationController
   include CourtCaseResponseHelper
-  REQUIRED_PARAMS = %i[id tenancy_ref court_date court_outcome balance_on_court_outcome_date].freeze
-  OPTIONAL_PARAMS = %i[strike_out_date terms disrepair_counter_claim].freeze
+  REQUIRED_UPDATE_PARAMS = %i[id tenancy_ref court_date court_outcome balance_on_court_outcome_date].freeze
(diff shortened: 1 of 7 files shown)


app/controllers/court_cases_controller.rb
```diff
@@ -14,16 +16,40 @@ def index
...
+    if updated_court_case
+      response = map_court_case_to_response(court_case: updated_court_case)
+      render json: response
+    else
+      render json: { error: 'court case not found' }, status: :not_found
+    end
+  end
+
+  def update_court_case_params
+    params.require(REQUIRED_UPDATE_PARAMS)
```
C-gyorfi, line 52: Does this line do anything in this case? I think `update_court_case_params` just returns line 53 🤔

lib/hackney/income/update_court_case.rb
```diff
@@ -0,0 +1,28 @@
...
+    class UpdateCourtCase
+      def execute(court_case_params:)
+        court_case_id = court_case_params[:id]
+        court_date = court_case_params[:court_date]
+        tenancy_ref = court_case_params[:tenancy_ref]
+
+        court_case = Hackney::Income::Models::CourtCase.find_by_id(court_case_id)
+
+        return if court_case.nil?
+        return if court_case.court_date != court_date
```
C-gyorfi, line 12: Isn't the court date editable?

app/controllers/court_cases_controller.rb
```diff
@@ -1,5 +1,7 @@
 class CourtCasesController < ApplicationController
   include CourtCaseResponseHelper
+  REQUIRED_UPDATE_PARAMS = %i[id tenancy_ref court_date court_outcome balance_on_court_outcome_date].freeze
+  OPTIONAL_UPDATE_PARAMS = %i[strike_out_date terms disrepair_counter_claim].freeze
```
C-gyorfi, line 4: Sorry if I'm being slow, but I don't really understand why we need required and optional params? Maybe we could just use define one constant with all the params?

spec/lib/hackney/income/update_court_case_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+
+  context 'when the court case does not exist' do
+    it 'returns nil' do
+      court_case_params[:id] = Faker::Number.number(digits: 6)
+
+      expect(subject.execute(court_case_params: court_case_params)).to be_nil
+    end
+  end
+
+  context 'when the court date of the existing court case does not match' do
```
C-gyorfi, line 72: Is the court case editable?

lib/hackney/income/update_court_case.rb
Ninamma: yup - realising all of this is actually preventing editing court dates so i think i will get rid of those and possibly also the required params and make it pretty much all optional in the back end since we will have required fields in the forms in the frontend 🤔

app/controllers/court_cases_controller.rb
Ninamma: i think this was meant to make sure those parameters are present, but yeah, if we're using this usecase to edit court dates then this doesn't make too much sense 🤔

app/controllers/court_cases_controller.rb
antonyoneill: If update is a PATCH then nothing should be required, except the id of the court case. tenancy_ref should be immutable

app/controllers/court_cases_controller.rb
```diff
@@ -14,16 +16,40 @@ def index
...
     new_court_case = income_use_case_factory.create_court_case.execute(court_case_params: court_case_params)
     response = map_court_case_to_response(court_case: new_court_case)
 
     render json: response
   end
+
+  def update
+    court_case_params = {
+      id: update_court_case_params[:id],
+      tenancy_ref: update_court_case_params[:tenancy_ref],
```
elena-vi, line 32: why do we care about the `tenancy_ref` here? 
Surely just `id` is enough find the relevant case

lib/hackney/income/update_court_case.rb
```diff
@@ -0,0 +1,28 @@
...
+        return if court_case.tenancy_ref != tenancy_ref
```
elena-vi, line 13: again, I'm a little bit confused by why we're double checking the tenancy_ref?

app/controllers/court_cases_controller.rb
antonyoneill: Think Ninamma had just not pushed :angel:

PR #470, opened 2020-08-11
Add ma court case helper

C-gyorfi:
## Context
We want to use MAA agreements rather than the agreements from UH whenever we running classification

- Add agreements helpers for MA court cases `MACourtCaseHelpers`

## Link to Jira card

## Things to check

### DIFF
--- a/lib/hackney/income/tenancy_classification/v2/helpers/ma_court_case_helpers.rb
+++ b/lib/hackney/income/tenancy_classification/v2/helpers/ma_court_case_helpers.rb
@@ -8,6 +8,7 @@ module Hackney
 
             def court_warrant_active?
               return false if most_recent_court_case.blank?
+              return false if most_recent_court_case.court_date.blank?
               return false if most_recent_court_case.court_outcome.blank?
 
               if most_recent_court_case.court_outcome.in?([
@@ -18,7 +19,7 @@ module Hackney
                 Hackney::Tenancy::UpdatedCourtOutcomeCodes::SUSPENSION_ON_TERMS,
                 Hackney::Tenancy::UpdatedCourtOutcomeCodes::STAY_OF_EXECUTION
               ])
-                return most_recent_court_case.court_date.blank? || most_recent_court_case.court_date > 6.years.ago
+                return most_recent_court_case.court_date > 6.years.ago
               end
 
               false
--- a/lib/hackney/income/tenancy_classification/v2/helpers/ma_court_case_helpers.rb
+++ b/lib/hackney/income/tenancy_classification/v2/helpers/ma_court_case_helpers.rb
@@ -11,12 +11,12 @@ module Hackney
               return false if most_recent_court_case.court_outcome.blank?
 
               if most_recent_court_case.court_outcome.in?([
-                Hackney::Tenancy::CourtOutcomeCodes::ADJOURNED_GENERALLY,
-                Hackney::Tenancy::CourtOutcomeCodes::ADJOURNED_ON_TERMS,
-                Hackney::Tenancy::CourtOutcomeCodes::ADJOURNED_ON_TERMS_SECONDARY,
-                Hackney::Tenancy::CourtOutcomeCodes::SUSPENDED_POSSESSION,
-                Hackney::Tenancy::CourtOutcomeCodes::OUTRIGHT_POSSESSION_FORTHWITH,
-                Hackney::Tenancy::CourtOutcomeCodes::OUTRIGHT_POSSESSION_WITH_DATE
+                Hackney::Tenancy::UpdatedCourtOutcomeCodes::ADJOURNED_GENERALLY_WITH_PERMISSION_TO_RESTORE,
+                Hackney::Tenancy::UpdatedCourtOutcomeCodes::ADJOURNED_TO_NEXT_OPEN_DATE,
+                Hackney::Tenancy::UpdatedCourtOutcomeCodes::ADJOURNED_TO_ANOTHER_HEARING_DATE,
+                Hackney::Tenancy::UpdatedCourtOutcomeCodes::ADJOURNED_FOR_DIRECTIONS_HEARING,
+                Hackney::Tenancy::UpdatedCourtOutcomeCodes::SUSPENSION_ON_TERMS,
+                Hackney::Tenancy::UpdatedCourtOutcomeCodes::STAY_OF_EXECUTION
               ])
                 return most_recent_court_case.court_date.blank? || most_recent_court_case.court_date > 6.years.ago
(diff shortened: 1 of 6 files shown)


spec/lib/hackney/income/tenancy_classification/v2/helpers/ma_court_case_helpers_spec.rb
```diff
@@ -0,0 +1,212 @@
...
+
+    context 'when there is a court date in the future' do
+      let(:most_recent_court_case) { { court_outcome: court_outcome, court_date: court_date } }
+      let(:court_date) { 2.days.ago }
+
+      context 'when the court outcome is blank' do
+        let(:court_outcome) { nil }
+
+        it 'returns false' do
+          expect(subject).to eq(true)
```
Ninamma, line 199: slightly confused about this bit because the description saying to return `false` and this bit in the code also says to return `false` if the date is in the future, right? 🤔 

I think this test was already like this - confused as to why it passes 

```
def court_outcome_missing?
              return false if court_date_in_future?
              return false if no_court_date?

              most_recent_court_case.court_outcome.blank?
            end
```

spec/lib/hackney/income/tenancy_classification/v2/helpers/ma_court_case_helpers_spec.rb
```diff
@@ -0,0 +1,212 @@
...
+        it 'returns false' do
+          expect(subject).to eq(false)
+        end
+      end
+
+      context 'with a court date is nil' do
+        let(:court_date) { nil }
+
+        it 'returns false' do
+          expect(subject).to eq(true)
```
Ninamma, line 147: also, am a bit unsure about this 🤔 - for MA I think we would want this to be false? 
I don't know though 😬

antonyoneill: The subject is `court_warrant_active?` - if there's no court_date we can't tell if there's an active court warrant so I think it should be false.

C-gyorfi: Good catch! This is a confusing one, `court_warrant_active?` when there is a court outcome, but no court date. We have the same weird test in the old helper_spec.

spec/lib/hackney/income/tenancy_classification/v2/helpers/ma_court_case_helpers_spec.rb
C-gyorfi: I think its expected to return true in any case when a court_outcome is nil `court_outcome_missing?` so its the test name that is incorrect, but I'm confused why would we return false if the court date in the future. 😕

C-gyorfi: I made some changes both in the existing helpers and in the new  MA helpers let me know if you agree with this behaviour: https://github.com/LBHackney-IT/lbh-income-api/pull/470/commits/40a5029fe92d09b5321bee0d87a9c61254d81f1b

C-gyorfi: I also had to change some test setups

lib/hackney/income/tenancy_classification/v2/helpers/ma_court_case_helpers.rb
```diff
@@ -34,8 +35,8 @@ def no_court_date?
             end
 
             def court_outcome_missing?
-              return false if court_date_in_future?
-              return false if no_court_date?
+              return true if court_date_in_future?
+              return true if no_court_date?
```
antonyoneill, line 40: I'm confused about this - if there is a court date in the future, the outcome can't be missing as there hasn't been a hearing, and therefore the absence of a court outcome is expected. Same for no court date

Should we chat about this?

lib/hackney/income/tenancy_classification/v2/helpers/uh_court_case_helpers.rb
```diff
@@ -30,8 +31,8 @@ def no_court_date?
             end
 
             def court_outcome_missing?
-              return false if court_date_in_future?
-              return false if no_court_date?
```
antonyoneill, line 34: as above

spec/lib/hackney/income/tenancy_classification/v2/helpers/helpers_spec.rb
```diff
@@ -512,8 +512,16 @@ def initialize(case_priority, criteria, documents)
     context 'when there is no court date' do
```
antonyoneill, line 512: as above, if there's no court date, then the court outcome cannot be classed as missing

spec/lib/hackney/income/tenancy_classification/v2/helpers/helpers_spec.rb
```diff
@@ -512,8 +512,16 @@ def initialize(case_priority, criteria, documents)
...
       let(:courtdate) { nil }
 
-      it 'returns false' do
-        expect(subject).to eq(false)
+      it 'returns true' do
+        expect(subject).to eq(true)
+      end
+    end
+
+    context 'when the court date is in future' do
```
antonyoneill, line 520: Likewise, if it's in the future, there has been no hearing and therefore cannot be missing a court outcome

spec/lib/hackney/income/tenancy_classification/v2/helpers/helpers_spec.rb
```diff
@@ -538,7 +546,7 @@ def initialize(case_priority, criteria, documents)
     end
 
     context 'when there is a court date in the future' do
-      let(:courtdate) { 2.days.ago }
```
antonyoneill, line 540: whoops. I clearly didn't pay attention as I wrote this test :(

spec/lib/hackney/income/tenancy_classification/v2/helpers/helpers_spec.rb
```diff
@@ -551,8 +559,8 @@ def initialize(case_priority, criteria, documents)
       context 'when the court outcome is not blank' do
         let(:court_outcome) { Hackney::Tenancy::CourtOutcomeCodes::OUTRIGHT_POSSESSION_WITH_DATE }
 
-        it 'returns false' do
```
antonyoneill, line 553: there's an outcome, with a court date in the future (this is an edge case of bad data from UH) and should return false, given that the court outcome is set (and therefore not missing)

PR #473, opened 2020-08-13
Add tighter check for adjourned? on court case model

C-gyorfi:
## Context
It can fail a test when faker generates court_outcome that starts with 'a'

- Add tighter check for adjourned? on court case model


antonyoneill:
I really don't think we should be using faker data in this way in general. We should be driving these tests with known court outcome codes.

PR #478, opened 2020-08-18
Add UH agreement gateway

antonyoneill:
## Context

We need to be able to sync UH agreements into the MAA database. In order to do so we need a solid way of extracting agreements out of UH.

This PR adds a gateway that allows us to extract all agreements from UH, we can later loop over each of these and create agreements within MAA.

## Link to Jira card

https://hackney.atlassian.net/browse/MAAP-224

## Things to check

### DIFF
--- a/lib/hackney/income/universal_housing_agreement_gateway.rb
+++ b/lib/hackney/income/universal_housing_agreement_gateway.rb
@@ -17,7 +17,7 @@ module Hackney
 
         results.map do |agreement|
           {
-            start_date: agreement[:arag_startdate],
+            start_date: agreement[:aragdet_startdate],
             breached: agreement[:arag_breached],
             starting_balance: agreement[:arag_startbal],
             comment: agreement[:aragdet_comment],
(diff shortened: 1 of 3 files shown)


lib/hackney/income/universal_housing_agreement_gateway.rb
```diff
@@ -0,0 +1,36 @@
...
+          WHERE arag.tag_ref = ?
+        SQL
+
+        results = @universal_housing_client[sql, tenancy_ref].all
+
+        results.map.with_index do |agreement, index|
+          final_row = index == results.count - 1
+          {
+            start_date: agreement[:aragdet_startdate],
+            status: final_row ? agreement[:arag_status] : Hackney::Income::CANCELLED_ARREARS_AGREEMENT_STATUS,
```
elena-vi, line 22: you might want to look at the codes on the front end in 
`lib/hackney/income/domain/arrears_agreement.rb`

PR #479, opened 2020-08-18
Informal agreement confirmation letter template

Miles-Alford:
## Context

When an informal agreement is created and entered into MAA a letter should be sent to the tenant who the agreement was made for, this PR adds this template into MAA 

- Add informal_agreement_confirmation_letter template

- Create InformalAgreement letter model to get the required data for the letter

-  Update Income Collection Letter model to be able to generate a InformalAgreement letter

## Guidance to review

Got some areas which I would like some feedback on, mainly on the InformalAgreement letter model itself. Would be good to know if the name of the attributes are clear enough and that the calculations for making the rent calculate by frequency are correct. 

Also would be good to know if getting the agreement on that class is correct, been thinking it may be good to add a select clause only for informal agreements.

## Link to Jira card

https://hackney.atlassian.net/browse/MAAP-370

## Things to check


app/models/hackney/income_collection/letter/informal_agreement.rb
```diff
@@ -0,0 +1,58 @@
...
+          @total_amount_payable = format('%.2f', calculate_total_amount_payable(@rent_charge, @instalment_amount))
+        end
+
+        private
+
+        def format_date(date)
+          date.strftime('%d %B %Y')
+        end
+
+        def find_rent(tenancy_ref)
```
elena-vi, line 28: let's avoid fetching things here, make weekly rent a mandatory param

app/models/hackney/income_collection/letter/informal_agreement.rb
```diff
@@ -0,0 +1,58 @@
...
+
+        def initialize(params)
+          super(params)
+
+          validated_params = validate_mandatory_fields(MANDATORY_FIELDS, params)
+          @agreement = find_agreement(validated_params[:tenancy_ref])
+          @rent = find_rent(validated_params[:tenancy_ref])
+          @rent_charge = format('%.2f', calculate_rent(@rent, @agreement.frequency))
+          @instalment_amount = @agreement.amount
+          @date_of_first_payment = format_date(@agreement.start_date)
```
elena-vi, line 18: let's make `instalment_amount` and `date_of_first_payment` and `agreement_frequency` as mandatory params

app/models/hackney/income_collection/letter.rb
```diff
@@ -8,12 +8,18 @@ class Letter
       attr_reader :tenancy_ref, :address_line1, :address_line2,
                   :address_line3, :address_line4, :address_post_code,
                   :payment_ref, :total_collectable_arrears_balance,
-                  :title, :forename, :surname, :errors, :tenant_address
+                  :title, :forename, :surname, :errors, :tenant_address, :rent,
+                  :agreement, :rent_charge, :total_amount_payable, :date_of_first_payment
```
elena-vi, line 12: let's move the agreement specific params into the agreements letter

app/models/hackney/income_collection/letter/informal_agreement.rb
```diff
@@ -0,0 +1,52 @@
...
+          @date_of_first_payment = validated_params[:date_of_first_payment]
+
+          return unless @rent
+          @rent_charge = format('%.2f', calculate_rent(@rent, @agreement_frequency))
+          @total_amount_payable = format('%.2f', calculate_total_amount_payable(@rent_charge, @instalment_amount))
+        end
+
+        private
+
+        def format_date(date)
```
C-gyorfi, line 28: We have a definition for this in `BeforeAction` letter, could define this in the parent class or even better move it in a date helper? Where do we call this atm? It could blow up if date is nil.

lib/hackney/pdf/templates/income/informal_agreement_confirmation_letter.erb
```diff
@@ -0,0 +1,58 @@
...
+    <strong>Rent Account Balance: </strong> <br>
+    <strong>Account Number: </strong>
+  </div>
+
+  <p>
+    <strong> Thank you for contacting me about your rent arrears, and this letter confirms the agreement you have made to clear your arrears.</strong>
+  </p>
+
+  <ul class = "no-bullets">
+    <li>Weekly/Monthly Rent: <%= @letter.rent_charge %> </li>
```
C-gyorfi, line 18: I'm not expecting you to change this, its more of a design question. Couldn't we pass down frequency to generate a clearer letter format?

spec/requests/agreements_spec.rb
```diff
@@ -155,7 +155,7 @@
 
           allow(Hackney::Income::CreateFormalAgreement).to receive(:new).and_return(create_formal_agreement_instance)
           allow(create_formal_agreement_instance).to receive(:execute)
-            .with(new_agreement_params: new_agreement_params)
+            .with(new_agreement_params: new_agreement_params.merge(frequency: frequency.to_sym))
```
C-gyorfi, line 158: Bit confused, why this change needed?

C-gyorfi (approved):
Picked on a few minor things but looks good

app/models/hackney/income_collection/letter/informal_agreement.rb
Miles-Alford: Good point, created a date helper so all letters should use that instead.

PR #482, opened 2020-08-18
Add uh court case migrator

antonyoneill:
## Context

We want to model court cases within MA. We're going to migrate the data in from UH on this so we can look up for formal agreements and allow officers to manage court cases through MA

Initial start on wiring up a court case migrator. It's responsibility is to determine whether it can/should create a court case in MA, and do so.

It will update court cases on subsequent syncs, provided there's only a single court case, and that it won't overwrite existing data on that single court case.

## Link to Jira card

https://hackney.atlassian.net/browse/MAAP-417

## Things to check

### DIFF
--- a/lib/hackney/income/migrate_uh_court_case.rb
+++ b/lib/hackney/income/migrate_uh_court_case.rb
@@ -8,7 +8,6 @@ module Hackney
       end
 
       def migrate(criteria)
-
         Rails.logger.debug { "Starting migration for UH court case for tenancy ref #{criteria.tenancy_ref}" }
 
         unless criteria_contains_court_data(criteria)
@@ -41,7 +40,7 @@ module Hackney
         court_case_params[:court_outcome] = nil if existing_court_case.court_outcome.present?
 
         if court_case_params.compact.keys.empty?
-          Rails.logger.debug { "UH Criteria does not contain any new information" }
+          Rails.logger.debug { 'UH Criteria does not contain any new information' }
           return
         end
 
@@ -55,14 +54,26 @@ module Hackney
       end
 
       def criteria_contains_court_data(criteria)
-        criteria.courtdate.present? || criteria.court_outcome.present?
+        !get_courtdate(criteria.courtdate).nil? || !map_court_outcome(criteria.court_outcome).nil?
       end
 
       def map_criteria_to_court_case_params(criteria)
         {
-          court_date: criteria.courtdate,
-          court_outcome: criteria.court_outcome
+          court_date: get_courtdate(criteria.courtdate),
+          court_outcome: map_court_outcome(criteria.court_outcome)
         }
+      end
+
+      def get_courtdate(courtdate)
+        return nil if courtdate == DateTime.parse('1900-01-01 00:00:00')
+
+        courtdate
+      end
+
+      def map_court_outcome(outcome)
+        return nil if outcome.strip.empty?
+
+        outcome.strip
       end
     end
   end
--- a/lib/hackney/income/migrate_uh_court_case.rb
+++ b/lib/hackney/income/migrate_uh_court_case.rb
@@ -1,17 +1,60 @@
 module Hackney
   module Income
     class MigrateUhCourtCase
-      def initialize(create_court_case:)
+      def initialize(create_court_case:, view_court_cases:, update_court_case:)
         @create_court_case = create_court_case
+        @view_court_cases = view_court_cases
+        @update_court_case = update_court_case
       end
 
       def migrate(criteria)
-        @create_court_case.execute(map_criteria_to_court_case_params(criteria)) if should_migrate_criteria(criteria)
+
+        Rails.logger.debug { "Starting migration for UH court case for tenancy ref #{criteria.tenancy_ref}" }
+
+        unless criteria_contains_court_data(criteria)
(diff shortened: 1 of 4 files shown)


lib/hackney/income/migrate_uh_court_case.rb
```diff
@@ -0,0 +1,69 @@
+module Hackney
+  module Income
+    class MigrateUhCourtCase
+      def initialize(create_court_case:, view_court_cases:, update_court_case:)
+        @create_court_case = create_court_case
+        @view_court_cases = view_court_cases
+        @update_court_case = update_court_case
+      end
+
+      def migrate(criteria)
```
elena-vi, line 10: is criteria the best variable name..?

elena-vi (approved):
this looks good.

I'm just wondering if we could classify any tenancies that have no court date but a court outcome 
as `check case data`, I think that will help the users to clean up the data

spec/lib/hackney/income/migrate_uh_court_case_spec.rb
```diff
@@ -0,0 +1,279 @@
...
+          )
+        )
+        subject
+      end
+    end
+
+    context 'when provided a criteria with a court outcome but without a court date' do
+      let(:criteria_attributes) {
+        {
+          court_outcome: Hackney::Tenancy::UpdatedCourtOutcomeCodes::ADJOURNED_FOR_DIRECTIONS_HEARING,
```
C-gyorfi, line 61: I'm sure you are aware of this and probably we should address this in another slice: UH court cases are in: `Hackney::Tenancy::CourtOutcomeCodes`  and we will need to map these when migrating.

spec/lib/hackney/income/migrate_uh_court_case_spec.rb
```diff
@@ -0,0 +1,279 @@
...
+          subject
+        end
+      end
+    end
+
+    context 'when a partial court case already exists without a court date but has an outcome' do
+      let(:existing_court_cases) {
+        [create(:court_case,
+                court_date: nil,
+                court_outcome: Hackney::Tenancy::CourtOutcomeCodes::ADJOURNED_GENERALLY)]
```
C-gyorfi, line 209: Maybe I was wrong in my previous comment, are the `Hackney::Tenancy::CourtOutcomeCodes` whet we meant to use in MA court_cases?

C-gyorfi (approved):
Nice work, I'm just a bit confused about the court outcomes

lib/hackney/income/migrate_uh_court_case.rb
antonyoneill: Good question! For the purposes of this, it's a legacy data record... That said we do get the criteria and call it that throughout the code base, and we use a `StubCriteria`. So I think this is best for now.

spec/lib/hackney/income/migrate_uh_court_case_spec.rb
antonyoneill: ! Thanks so much. I forgot about mapping them! :disappointed:

spec/lib/hackney/income/migrate_uh_court_case_spec.rb
antonyoneill: I actually got a bit confused, there's two:
- `Hackney::Tenancy::CourtOutcomeCodes`
- `Hackney::Tenancy::UpdatedCourtOutcomeCodes`

I wasn't really sure what it meant..

antonyoneill:
@elena-vi wrt to the check case data, we do already have that pointing to the old UH data. We should update/create a new MA ruleset like Csaba did for the agreement breach rulesets

lib/hackney/income/migrate_uh_court_case.rb
elena-vi: I like the idea of phasing out the term `criteria` on the leasehold side I've started calling the thing that we get from UH as [`attributes`](https://github.com/LBHackney-IT/lbh-income-api/blob/master/lib/hackney/leasehold/sync_action_attributes.rb#L10)

spec/lib/hackney/income/migrate_uh_court_case_spec.rb
```diff
@@ -21,12 +21,15 @@
   let(:criteria) { Stubs::StubCriteria.new(criteria_attributes) }
   let(:existing_court_cases) { [] }
 
+  UH_NIL_COURT_OUTCOME = '   '.freeze
+  UH_NIL_DATE = DateTime.parse('1900-01-01 00:00:00')
```
C-gyorfi, line 25: Ugh, maybe a lame question but why though 🤔

antonyoneill: Why does UH do this? 
- It was fairly normal "back in the old days" to reserve the data space that you would potentially occupy for performance reasons. ie the `u_court_outcome` is defined as a 3 character varchar, so they'd pad it with spaces.. :eyes: 
- It's also likely that the form in UHW that they'd input the 3 characters would be prepopulated with spaces, and you'd replace over it - think that this system was probably designed and built by people that used to design the old terminal applications like
  image

Why do I have to do this?
- I don't want to create court cases in MA with the date at the beginning of the century, with a court outcome of `"   "` :joy:

C-gyorfi: Thanks for explaining 🤓 👍

PR #491, opened 2020-08-24
Fix migrate uh court case

elena-vi:
## Context

using the migrate uh court case usecae was causing errors

changed the way params are passed into update/create court case usecases

## Things to check


lib/hackney/income/migrate_uh_court_case.rb
```diff
@@ -39,12 +39,12 @@ def migrate(criteria)
         court_case_params[:court_date] = nil if existing_court_case.court_date.present?
         court_case_params[:court_outcome] = nil if existing_court_case.court_outcome.present?
 
-        if court_case_params.compact.keys.empty?
+        if court_case_params[:court_date].nil? || court_case_params[:court_outcome].nil?
```
Ninamma, line 42:
```suggestion
        if court_case_params[:court_date].nil? && court_case_params[:court_outcome].nil?
```
I might be being very silly here, but should this be an AND? 🤔

elena-vi: yes, you are correct!

PR #495, opened 2020-08-25
MERGE WHEN RELEASING AGREEMENTS - Enable using MA agreements and court cases data in classifier

C-gyorfi:
## Context
We are using agreements data that is synced from UH when running classifier, going forward we want to use MA agreement system as the data source.

- Enable using MA agreements and court cases data in classifier


Ninamma (approved):
😍 that's so cool

PR #496, opened 2020-08-25
migrate uh agreement usecase

elena-vi:
## Context

- when there is a UH agreement and a court case we create a formal agreement
- when there are migrated agreements in the system, we ignore UH agreements

## Things to check

### DIFF
--- a/lib/hackney/income/migrate_uh_agreement.rb
+++ b/lib/hackney/income/migrate_uh_agreement.rb
@@ -25,28 +25,41 @@ module Hackney
 
         return unless uh_agreements.any?
 
-        ma_agreemnets = @view_agreements.execute(tenancy_ref: tenancy_ref)
+        ma_agreements = @view_agreements.execute(tenancy_ref: tenancy_ref)
 
-        return unless ma_agreemnets.empty?
+        return unless ma_agreements.empty?
 
         court_cases = @view_court_cases.execute(tenancy_ref: tenancy_ref)
 
         if court_cases.any?
-          last_agreement = uh_agreements.pop
+          formal_agreement = uh_agreements.pop
 
-          new_agreement = create_formal_agreement(tenancy_ref, last_agreement, court_cases.last.id)
+          uh_agreements.each do |agreement|
+            migrate_informal_agreement(agreement, tenancy_ref)
+          end
 
-          add_agreement_migration(last_agreement[:uh_id], new_agreement.id)
-        end
+          migrate_formal_agreement(court_cases, formal_agreement, tenancy_ref)
 
-        uh_agreements.each do |agreement|
-          new_agreement = create_informal_agreement(tenancy_ref, agreement)
-
-          add_agreement_migration(agreement[:uh_id], new_agreement.id)
+        else
+          uh_agreements.each do |agreement|
+            migrate_informal_agreement(agreement, tenancy_ref)
+          end
         end
       end
 
       private
+
+      def migrate_formal_agreement(court_cases, formal_agreement, tenancy_ref)
+        new_agreement = create_formal_agreement(tenancy_ref, formal_agreement, court_cases.last.id)
+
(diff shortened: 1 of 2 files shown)


lib/hackney/income/migrate_uh_agreement.rb
```diff
@@ -0,0 +1,105 @@
...
+
+        if court_cases.any?
+          last_agreement = uh_agreements.pop
+
+          new_agreement = create_formal_agreement(tenancy_ref, last_agreement, court_cases.last.id)
+
+          add_agreement_migration(last_agreement[:uh_id], new_agreement.id)
+        end
+
+        uh_agreements.each do |agreement|
```
C-gyorfi, line 42: Why are we generating informal agreements after the formal created? Since we only allow one active agreement at a time, we cancel out previous agreements, however, you should not be allowed to create an informal agreement after a formal. *this is something we should address in the front-end too

```RUBY
        if active_agreements.any?
          raise CreateAgreementError, 'There is an existing formal agreement for this tenancy' if active_agreements.any?(&:formal?)

          cancel_active_agreements(active_agreements)
        end
```

elena-vi: the idea here is if there are multiple agreements in UH, and there is a court date. The last agreement will be made formal and the others are informal. I'll change the order of how these are created

spec/lib/hackney/income/migrate_uh_agreement_spec.rb
```diff
@@ -0,0 +1,195 @@
...
+        uh_id: uh_id
+      }]
+    }
+
+    before do
+      expect(view_agreements).to receive(:execute).and_return([])
+      expect(view_court_cases).to receive(:execute).and_return([court_case])
+    end
+
+    it 'migrates an informal agreement' do
```
Ninamma, line 168:
```suggestion
    it 'migrates a formal agreement' do
```
bit nitpicky 😅  but it's this right? 🤔

lib/hackney/income/migrate_uh_agreement.rb
```diff
@@ -0,0 +1,105 @@
...
+        uh_agreements = Hackney::UniversalHousing::Client.with_connection do |database|
+          Hackney::Income::UniversalHousingAgreementGateway.for_tenancy(
+            database,
+            tenancy_ref
+          )
+        end
+
+        return unless uh_agreements.any?
+
+        ma_agreemnets = @view_agreements.execute(tenancy_ref: tenancy_ref)
```
Ninamma, line 28:
```suggestion
        ma_agreements = @view_agreements.execute(tenancy_ref: tenancy_ref)
```
teeny typo

lib/hackney/income/migrate_uh_agreement.rb
```diff
@@ -0,0 +1,105 @@
...
+
+        return unless ma_agreemnets.empty?
```
Ninamma, line 30:
```suggestion
        return unless ma_agreements.empty?
```
same as above

spec/lib/hackney/income/migrate_uh_agreement_spec.rb
```diff
@@ -0,0 +1,195 @@
...
+
+      expect(create_agreement_migration).to receive(:execute).with(
+        agreement_migration_params: {
+          agreement_id: new_agreement.id,
+          legacy_id: uh_id
+        }
+      )
+
+      subject.migrate(tenancy_ref: tenancy_ref)
+    end
```
C-gyorfi, line 193: Can we try?

`context 'when there are multiple UH agreements, no MA agreements and a court case'`

 I may be wrong but I reckon it would blow up

lib/hackney/income/migrate_uh_agreement.rb
C-gyorfi: Make sense 👍

Ninamma (approved):
👍

PR #499, opened 2020-08-27
Enable agreements migration

elena-vi:
## Context

Time to migrate these agreements when we sync every night! 🌔

- Call migrate_uh_agreement usecase when syncing case priorities
- make migrate_uh_agreement use the generic create agreement usecase

## Things to check

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -30,7 +30,7 @@ Metrics/ClassLength:
   Enabled: false
 
 Metrics/LineLength:
-  Max: 160
+  Max: 180
 
 Metrics/MethodLength:
   Enabled: false
--- a/lib/hackney/income/use_case_factory.rb
+++ b/lib/hackney/income/use_case_factory.rb
@@ -194,9 +194,15 @@ module Hackney
         Hackney::Income::MigrateUhAgreement.new(
           view_agreements: view_agreements,
           view_court_cases: view_court_cases,
-          create_informal_agreement: create_informal_agreement,
-          create_formal_agreement: create_formal_agreement,
+          create_agreement: create_agreement,
           create_agreement_migration: create_agreement_migration
+        )
+      end
+
+      def create_agreement
+        Hackney::Income::CreateAgreement.new(
+          add_action_diary: add_action_diary,
+          cancel_agreement: cancel_agreement
         )
       end
 
--- a/lib/hackney/income/use_case_factory.rb
+++ b/lib/hackney/income/use_case_factory.rb
@@ -185,8 +185,23 @@ module Hackney
           prioritisation_gateway: prioritisation_gateway,
           stored_worktray_item_gateway: stored_worktray_item_gateway,
           update_agreement_state: update_agreement_state,
-          migrate_court_case_usecase: migrate_court_case_usecase
+          migrate_court_case_usecase: migrate_court_case_usecase,
+          migrate_uh_agreement: migrate_uh_agreement
         )
+      end
+
+      def migrate_uh_agreement
+        Hackney::Income::MigrateUhAgreement.new(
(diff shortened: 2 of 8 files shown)


spec/lib/hackney/income/migrate_uh_agreement_spec.rb
```diff
@@ -35,8 +33,8 @@
     it 'does not call anything' do
       expect(view_agreements).not_to receive(:execute)
       expect(view_court_cases).not_to receive(:execute)
-      expect(create_informal_agreement).not_to receive(:execute)
-      expect(create_formal_agreement).not_to receive(:execute)
+      expect(create_agreement).not_to receive(:execute)
+      expect(create_agreement).not_to receive(:execute)
```
Ninamma, line 37:
```suggestion
      expect(create_agreement).not_to receive(:execute)
```
slightly nitpicky but also i don't know rspec too well, do we need both of these here? 🤔

C-gyorfi: I assume not, but line 37 is redundant

C-gyorfi:
```RUBY
CreateInformalAgreement < CreateAgreement
```
Yep CreateInformalAgreement/CreateFormalAgreement is a `CreateAgreement`

.rubocop.yml
```diff
@@ -30,7 +30,7 @@ Metrics/ClassLength:
   Enabled: false
 
 Metrics/LineLength:
-  Max: 160
+  Max: 180
```
C-gyorfi, line 33: Do we even need this then?

spec/lib/hackney/income/migrate_uh_agreement_spec.rb
```diff
@@ -72,8 +69,8 @@
 
     it 'nothing get\'s migrated' do
       expect(view_court_cases).not_to receive(:execute)
-      expect(create_informal_agreement).not_to receive(:execute)
-      expect(create_formal_agreement).not_to receive(:execute)
+      expect(create_agreement).not_to receive(:execute)
```
Ninamma, line 72: wait, create agreement use case doesn't have the `execute` method at all right? 🤔 so 
```suggestion
      expect(create_agreement).not_to receive(:create_agreement)
```
here and everywhere else we're testing for `execute` ? 🤔

.rubocop.yml
elena-vi: yes for [this line](https://github.com/LBHackney-IT/lbh-income-api/pull/499/files#diff-7ae5ae1dd936292410260669daf0c4a2R8)

lib/hackney/income/use_case_factory.rb
```diff
@@ -185,10 +185,31 @@ def sync_case_priority
...
+        Hackney::Income::MigrateUhAgreement.new(
+          view_agreements: view_agreements,
+          view_court_cases: view_court_cases,
+          create_agreement: create_agreement,
+          create_agreement_migration: create_agreement_migration
+        )
+      end
+
+      def create_agreement
+        Hackney::Income::CreateAgreement.new(
```
C-gyorfi, line 203: I know this concept doesn't really exist in rails, but this meant to be an abstract class, the problem with using this class for creating new agreements is it won't cancel previous agreements, the cancellation rules are implemented in the inherited classes because there is a hierarchy between Informal and Formal agreements.

.rubocop.yml
C-gyorfi: I mean if we don't want to follow rubocop style suggestions we could remove the rule itself, alternatively break the initialiser in multiple lines for readability?

C-gyorfi (approved):
It looks good

PR #504, opened 2020-08-28
Fix for when there's not court date (from UH)

Ninamma:
## Context

We run into an error after migrating agreements and court cases from UH.

[Court date is nil when checking for breaches](https://sentry.io/organizations/london-borough-of-hackney/issues/1863378649/?project=1276456&referrer=slack) 

Check if court date is present before comparing to see if it has passed


C-gyorfi (approved):
👍

elena-vi (approved):

PR #508, opened 2020-09-02
We want to add action diary entries when sending agreement letters

C-gyorfi:
## Context 
We want to add action diary entries when sending agreement letters to track last communication action and generate the
correct next recommended actions. Because of the missing action code, sending confirmation letter blows up:
https://sentry.io/organizations/london-borough-of-hackney/issues/1872201316/?project=1276456&referrer=slack

- Add missing action codes and for agreement confirmation letters
- Use the old action codes for agreement breach letters

## Things to check


Ninamma (approved):
🙌

PR #509, opened 2020-09-02
Remove unused or depreciated agreement fields

C-gyorfi:
## Context
There are many agreement related fields pulled when syncing case priority from UH. These fields are also getting passed to the front-end when calling `ViewCases` use case. Going forward we want to use agreement data from MA. 

- Remove references of unused/depreciated agreements fields 🧹  🗑️ 

## Things to check


spec/lib/hackney/income/view_cases_spec.rb
```diff
@@ -130,14 +119,8 @@
                                              current_arrears_agreement_status: nil,
 
                                              # TODO: the following fields needs update, no longer should be pulled from UH agreements
```
antonyoneill, line 121: Can these comments be removed now?

spec/lib/hackney/income/view_cases_spec.rb
```diff
@@ -154,16 +137,9 @@
           expect(subject.cases).to include(a_hash_including(
                                              ref: tenancy_attributes.fetch(:ref),
                                              current_arrears_agreement_status: agreements.last.current_state,
-
                                              # TODO: the following fields needs update, no longer should be pulled from UH agreements
```
antonyoneill, line 140: Can these comments be removed now?

antonyoneill (approved):
This looks OK

I think `broken_court_order` was being serialised in the tenancies_controller before hand, did you check this is not being used in the UI? (I suspect you have, just want to double check)

spec/lib/hackney/income/view_cases_spec.rb
C-gyorfi: Working on this, I just wanted to finish my day with a PR :D

PR #512, opened 2020-09-03
Fix court case fields on CasePriority

C-gyorfi:
## Context
There are `court_case` and `court_date` fields on `CasePriority` model, which is currently synced from UH. To show accurate data to the users, we want to show the `CourtCase` data that lives and managed in MA. 

- Change `CourtCase` to update related `CasePriority` whenever a court case created or changed
- Use court case data from MA when syncing case priorities 

## Guidance to review
I went down a miserable route and tried to remove `courtdate` and `court_outcome` fields from `CasePriority` model, but these are used when filtering tenancies, so we'd need to make changes in `WorktrayItemGateway` to be able to filter tenancies by `Upcoming Court Dates`.

By updating `CasePriority` whenever a court case created allows us to preserve the existing behaviour with the minimum change, also better for optimisation when filtering by `court_date` 

## Things to check

### DIFF
--- a/app/models/hackney/income/models/court_case.rb
+++ b/app/models/hackney/income/models/court_case.rb
@@ -2,6 +2,9 @@ module Hackney
   module Income
     module Models
       class CourtCase < ApplicationRecord
+        after_create :update_case_priority
+        after_update :update_case_priority
+
         validates_presence_of :tenancy_ref
         validates_inclusion_of :terms, in: [true, false], if: :can_have_terms?
         validates_inclusion_of :disrepair_counter_claim, in: [true, false], if: :can_have_terms?
@@ -38,6 +41,10 @@ module Hackney
 
           valid_court_outcomes = COURT_OUTCOMES_THAT_CAN_HAVE_TERMS + OTHER_COURT_OUTCOMES
           errors.add(:court_outcome, 'must be a valid court outcome code') unless valid_court_outcomes.include?(court_outcome)
+        end
+
+        def update_case_priority
+          Hackney::Income::Models::CasePriority.find_by(tenancy_ref: tenancy_ref)&.update!(court_outcome: court_outcome, courtdate: court_date)
         end
       end
     end
--- a/app/models/hackney/income/models/court_case.rb
+++ b/app/models/hackney/income/models/court_case.rb
@@ -2,8 +2,8 @@ module Hackney
   module Income
     module Models
       class CourtCase < ApplicationRecord
-        after_create :update_case_priority
-        after_update :update_case_priority
+        after_create :update_associated_worktray_item
+        after_update :update_associated_worktray_item
 
         validates_presence_of :tenancy_ref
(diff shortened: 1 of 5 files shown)


lib/hackney/income/worktray_item_gateway.rb
```diff
@@ -2,9 +2,11 @@ module Hackney
   module Income
     class WorktrayItemGateway
       GatewayModel = Hackney::Income::Models::CasePriority
+      CourtCaseModel = Hackney::Income::Models::CourtCase
 
       def store_worktray_item(tenancy_ref:, criteria:, classification:)
         gateway_model_instance = GatewayModel.find_or_initialize_by(tenancy_ref: tenancy_ref)
+        court_case = CourtCaseModel.where(tenancy_ref: tenancy_ref).last
```
antonyoneill, line 9: :+1:

spec/models/hackney/income/models/court_cases_spec.rb
```diff
@@ -47,6 +47,39 @@
...
+    let(:case_priority) { create(:case_priority, tenancy_ref: tenancy_ref, courtdate: nil, court_outcome: nil) }
+
+    it 'updates the court case details on the existing case priority' do
+      expected_court_date = Faker::Date.between(from: 10.days.ago, to: 3.days.ago)
+      expected_court_outcome = valid_outcomes_without_terms
+
+      expect(case_priority.courtdate).to be_nil
+      expect(case_priority.court_outcome).to be_nil
+
+      court_case = described_class.create!(
```
antonyoneill, line 60: What's the `!` for here? All the other `described_class.create` invocations don't use the bang

app/models/hackney/income/models/court_case.rb
```diff
@@ -39,6 +42,10 @@ def court_outcome_is_valid
           valid_court_outcomes = COURT_OUTCOMES_THAT_CAN_HAVE_TERMS + OTHER_COURT_OUTCOMES
           errors.add(:court_outcome, 'must be a valid court outcome code') unless valid_court_outcomes.include?(court_outcome)
         end
+
+        def update_case_priority
+          Hackney::Income::Models::CasePriority.find_by(tenancy_ref: tenancy_ref)&.update!(court_outcome: court_outcome, courtdate: court_date)
```
antonyoneill, line 47: Worth a comment here as to why you need to do this?

antonyoneill (approved):

spec/models/hackney/income/models/court_cases_spec.rb
C-gyorfi: `!` will blow up if fails (e.g failed validation), create fails quietly and returns nil

app/models/hackney/income/models/court_case.rb
C-gyorfi: Not a fan of comments(code smell?), do you think it's unclear from the implementation?

PR #513, opened 2020-09-03
when uh frequency is not in MA, write a comment about what it was

elena-vi:
## Context

UH is weird and wonderful and agreements have all sorts of frequencies, and we only support half of them. 

The following PR depends on https://github.com/LBHackney-IT/lbh-income-api/pull/515 being merged in

- when an agreement with an unsupported frequency is migrated, we set the frequency to [`unsupported_legacy_frequency `](https://github.com/LBHackney-IT/lbh-income-api/pull/515) and write a comment about what the frequency used to be 
- when a live agreement with an unsupported frequency, we raise a error what will be captured by sentry so that officers can amend the case accordingly in UH before we sync 

## Link to Jira card

https://hackney.atlassian.net/secure/RapidBoard.jspa?rapidView=30&modal=detail&selectedIssue=MAAP-469
## Things to check

### DIFF
--- a/lib/hackney/income/migrate_uh_agreement.rb
+++ b/lib/hackney/income/migrate_uh_agreement.rb
@@ -1,6 +1,8 @@
 module Hackney
   module Income
     class MigrateUhAgreement
+      MigrateUhAgreementError = Class.new(StandardError)
+
       def initialize(
         view_agreements:,
         view_court_cases:,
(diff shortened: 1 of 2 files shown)


antonyoneill:
Does this mean we'll have agreements in the database without a frequency? Is that allowed? :/

elena-vi:
We've discussed this at stand up and agreed that it was okay, also Ian is going to make sure that the current live agreements with unsupported frequencies are amended on UH

antonyoneill:
I'm kinda surprised to see frequency isn't a required field on the Agreement model.

When I say 'allowed' I mean, should we be allowing inconsistent data to be stored in our database?

Does that have implications on breach detection, and how we render the agreement in the UI?

antonyoneill:
Looking at breach detection, it'll default to a month frequency if it's not "weekly", "fortnightly", "4 weekly" :+1:

antonyoneill (approved):

antonyoneill:
Great, thanks El!

PR #516, opened 2020-09-07
Update PR template

C-gyorfi:
## Context
Last Friday we had a minor outage when deployed a DB migration that dropped a field from DB. It's generally a good practice not to deploy a DB migration and code that relies on this migration in the same PR. 

- Update PR template with guidance for DB migration


Ninamma (approved):
🚓 👍

PR #521, opened 2020-09-08
Use courtoutcome balance for formal agreements

elena-vi:
## Context

When a formal agreement is created we need to make sure that the starting balance is the same as the balance at the court outcome

use the court outcome balance when creating a formal agreement

## Things to check


C-gyorfi (approved):
LGTM

PR #526, opened 2020-09-15
Allow creating variable and one-off payment agreement confirmation letters

C-gyorfi:
## Context
As a caseworker, I want to be able to send agreement confirmation letters for different payment conditions, so that I can give clear guidance to the tenant how much they need to pay.

- Allow creating variable payment agreement confirmation letter
- Allow creating one-off payment agreement confirmation letter

### Templates
image

image

## Link to Jira card
https://hackney.atlassian.net/browse/MAAP-488

## Things to check


elena-vi (approved):
This looks good, any idea how court outcome with terms should look like?

PR #532, opened 2020-09-16
Change wording for date of first payment

C-gyorfi:
## Context
Ian requested a change on the wording of the `date of first payment`

- Use `date of first (monthly/weekly/etc.) payment: ` so its clear that this refers to the regular payments when there is a lump sum payment


Ninamma (approved):
👍

PR #534, opened 2020-09-16
Call update_agreement_state when creating formal/informal agreements

C-gyorfi:
## Context
Elaine raised that entering historical agreements can cause confusions. Because the breach detector runs nightly, even if these agreements meant to breached according to the entered data, they would not be breached until next day morning.
Would be a better user experience if agreements brought up to date when created.

- Call update_agreement_state when creating formal/informal agreements

## Things to check


Ninamma (approved):
👍 🎉

PR #536, opened 2020-09-17
Add GET eviction dates endpoint

Miles-Alford:
## Context

This endpoint will allow users to view eviction dates that have been added onto a tenancy

- Add GET create eviction date endpoint to allow an eviction dates to be viewed
- Add ViewEvictionDates usecase
- Update API docs

## Guidance to review

Again pretty similar to how the court cases implementation, unsure if its possible for a tenancy to have multiple eviction dates but followed how it is done for agreements and court cases
## Link to Jira card

https://hackney.atlassian.net/browse/MAAP-491
## Things to check

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -36,6 +36,7 @@ Rails.application.routes.draw do
     patch '/court_case/:id/update', to: 'court_cases#update'
 
     post '/eviction_date/:tenancy_ref', to: 'eviction_dates#create'
+    get '/eviction_dates/:tenancy_ref', to: 'eviction_dates#index'
 
     get 'actions', to: 'actions#index'
   end
--- a/docs/api/v1/api.yaml
+++ b/docs/api/v1/api.yaml
@@ -172,6 +172,32 @@ paths:
       tags:
         - court_cases
 
+
+  /eviction_dates/{tenancy_ref}:
+    get:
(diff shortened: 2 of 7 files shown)


config/routes.rb
```diff
@@ -36,6 +36,7 @@
     patch '/court_case/:id/update', to: 'court_cases#update'
 
     post '/eviction_date/:tenancy_ref', to: 'eviction_dates#create'
+    get '/eviction_dates/:tenancy_ref', to: 'eviction_dates#index'
```
antonyoneill, line 39: Hey, Just noticed this - should it be `/eviction_date/` ?

docs/api/v1/api.yaml
```diff
@@ -172,6 +172,32 @@ paths:
       tags:
         - court_cases
 
+
+  /eviction_dates/{tenancy_ref}:
```
antonyoneill, line 176: date ?

antonyoneill (changes requested):
Given that we'd expect to overwrite the eviction date, I expect that there's only ever a single eviction date for a tenancy_ref?

Also would then expect only a single evictionDate to be sent back in the response rather than a list?

antonyoneill (approved):
Given the chat we just had about maintaining a history of eviction dates for a case I'm happy! :)

PR #349, opened 2020-01-29
instead of syncing pause case until tomorrow

elena-vi:

PR #538, opened 2020-09-21
resync after action diary entry

elena-vi:
## Context

as we have switched to live UH, we can successfully resync after an action and see a difference on production 

reversing https://github.com/LBHackney-IT/lbh-income-api/pull/349

## Link to Jira card

https://hackney.atlassian.net/browse/MAAP-414
## Things to check


lib/hackney/notification/send_manual_precompiled_letter.rb
```diff
@@ -9,14 +9,14 @@ def execute(username: nil, payment_ref: nil, template_id:, unique_reference:, le
           )
 
         # FIXME: this must be in a background job => UH is unreliable
-        # TODO: create job to accept exact same args as add_action_diary_and_pause_case_usecase
+        # TODO: create job to accept exact same args as add_action_diary_and_sync_case_usecase
```
C-gyorfi, line 12: Not related to the PR but I wonder how do we need to fix this? FIXME comments make me sad

C-gyorfi (approved):
👍

PR #539, opened 2020-09-21
Update eviction date table

Miles-Alford:
## Context

The eviction date table within the database is named poorly as it is called `eviction_dates` and contains a field called `eviction_date`. This PR renames the this table to be much clearer, renaming both the name of the table and the field.

- Change `eviction_dates` table & `eviction_date` field to `evictions` & `date`
- Update all references of eviction date to eviction: model, usecases, controller, tests
- Update `migrate_uh_eviction_date` to pass in the correct params to the `create_agreement` usecase

## Guidance to review

I think my main concern was the renaming of the `migrate_uh_eviction_date` usecase to `migrate_uh_eviction`, did it so that it is associated with the renamed eviction data model just in case someone new possibly goes search for a eviction date model
## Things to check


Ninamma (approved):
👍 although the word `eviction` doesn't look real to me any more

PR #541, opened 2020-09-22
migrate eviction dates at sync

elena-vi:
## Context

migrating the eviction date from uh into MA at sync 

## Link to Jira card

https://hackney.atlassian.net/browse/MAAP-492

## Things to check


Ninamma (approved):
👍

PR #547, opened 2020-09-24
Write to action diary and sync when court case created

Miles-Alford:
## Context

When a court case is created it should be written to the action diary that one has been created and should re-sync the case

This is the first slice

- Add `add_action_diary_and_sync_case` usecase to `create_court_case` usecase to allow adding to the action diary and re-syncing when a case is created

- `court_cases_controller` now gets the username from the request

## Guidance to review

The action diary message is quite basic, is it missing anything?
## Link to Jira card

https://hackney.atlassian.net/browse/MAAP-521
## Things to check

### DIFF
--- a/app/controllers/court_cases_controller.rb
+++ b/app/controllers/court_cases_controller.rb
@@ -23,11 +23,10 @@ class CourtCasesController < ApplicationController
       balance_on_court_outcome_date: create_court_case_params[:balance_on_court_outcome_date],
       strike_out_date: create_court_case_params[:strike_out_date],
       terms: create_court_case_params[:terms],
-      disrepair_counter_claim: create_court_case_params[:disrepair_counter_claim],
-      username: create_court_case_params[:username]
+      disrepair_counter_claim: create_court_case_params[:disrepair_counter_claim]
     }
 
-    new_court_case = income_use_case_factory.create_court_case.execute(court_case_params: court_case_params)
+    new_court_case = income_use_case_factory.create_court_case_and_sync.execute(court_case_params: court_case_params, username: create_court_case_params[:username])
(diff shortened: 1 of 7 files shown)


spec/requests/court_cases_spec.rb
```diff
@@ -21,6 +21,7 @@
     path '/court_case/{tenancy_ref}' do
       context 'when creating a new court case by adding a court date'
       let(:create_court_case_instance) { instance_double(Hackney::Income::CreateCourtCase) }
+      let(:username) { Faker::Name.name }
```
Ninamma, line 24: not 100% necessary, but guessing this will go in before my changes, and I also need `username` so it could be nice to have it declared outside this describe block? 😅

app/controllers/court_cases_controller.rb
```diff
@@ -12,7 +12,7 @@ def index
   end
 
   def create
-    parameters = %i[tenancy_ref court_date court_outcome balance_on_court_outcome_date strike_out_date terms disrepair_counter_claim].freeze
+    parameters = %i[tenancy_ref court_date court_outcome balance_on_court_outcome_date strike_out_date terms disrepair_counter_claim username].freeze
```
antonyoneill, line 15: ~Are you expecting the frontend to pass you the username in the data? Can we not get it from their token?~

Ignore me, I suspect you'll be passing this from the frontend using the username they are logged in as, from their token.

antonyoneill (approved):
Great!

app/controllers/court_cases_controller.rb
Miles-Alford: yep, the pull request for that: https://github.com/LBHackney-IT/LBH-IncomeCollection/pull/393

PR #549, opened 2020-09-24
Tighten some of the classification rules

C-gyorfi:
## Context
Noticed while fixing a classification bug, that informal agreement breach actions could have a tighter check

- Check if the agreement is informal in `informal agreement breach?` helper
- Show `:check_data` action when the court agreement is missing


spec/lib/hackney/income/tenancy_classification/helpers/agreement_helpers_spec.rb
```diff
@@ -135,10 +135,9 @@ def initialize(case_priority, criteria, documents)
     end
 
     context 'when agreemnt is not court ordered and it breached' do
-      it 'returns false' do
```
antonyoneill, line 138: I remember there was a lot of confusion around these tests when we originally implemented them. This looks right to me now! :)

antonyoneill (approved):

PR #550, opened 2020-09-25
add action diary and sync after adding eviction date

elena-vi:
## Context

we need to resync the case when an eviction date is added so that the case could be reclassified 

added a new usecase that will save the date, add the action diary and sync the case

also fixed the eviction date migration bug https://sentry.io/organizations/london-borough-of-hackney/issues/1910471839/?project=1276456&query=is%3Aunresolved

## Link to Jira card

https://hackney.atlassian.net/secure/RapidBoard.jspa?rapidView=30&modal=detail&selectedIssue=MAAP-523

## Things to check

### DIFF
--- a/lib/hackney/income/migrate_uh_eviction.rb
+++ b/lib/hackney/income/migrate_uh_eviction.rb
@@ -27,7 +27,7 @@ module Hackney
           return
         end
 
-        return unless uh_eviction.present? && existing_evictions.last.eviction_date < uh_eviction
+        return unless uh_eviction.present? && existing_evictions.last.date < uh_eviction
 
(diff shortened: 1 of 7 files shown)


antonyoneill (approved):

lib/hackney/income/migrate_uh_eviction.rb
```diff
@@ -27,7 +27,7 @@ def migrate(criteria)
           return
         end
 
-        return unless uh_eviction.present? && existing_evictions.last.eviction_date < uh_eviction
+        return unless uh_eviction.present? && existing_evictions.last.date < uh_eviction
```
antonyoneill, line 30: :+1: this is causing https://sentry.io/organizations/london-borough-of-hackney/issues/1910471839/?project=1276456&query=is%3Aunresolved

PR #551, opened 2020-09-25
Support unmapped uh frequency codes

antonyoneill:
## Context

There are some codes in UH that we haven't got mapped - while we should aim to add the unmapped ones, we should also be able to handle those that we do not have.

Use a case statement instead of a map, and have a default for those unknown ones.

## Link to Jira card

https://sentry.io/organizations/london-borough-of-hackney/issues/1883520536/?project=1276456&query=is%3Aunresolved

## Things to check


C-gyorfi (approved):
👍

PR #553, opened 2020-09-25
Add action diary entry and sync when adding a court outcome

Ninamma:
## Context

When an officer adds a court outcome, we want an entry to be added to the action diary and the case to be synced.

- Created new UpdateCourtCaseAndSync use case
   - Implements the UpdateCourtCase and AddActionDiaryAndSyncCase
   - This was created because those two use cases have a circular dependency, which this new use case avoids

- Make Court Case Controller use this new use case for the Court Case Update endpoint
## Guidance to review

There is some refactoring to be done court_cases_spec.rb, which I will try and do now.

I'm still not 100% confident when it comes to RSpec and test doubles so please shout if my use of them in the two tests doesn't make sense!

## Link to Jira card

https://hackney.atlassian.net/browse/MAAP-522?atlOrigin=eyJpIjoiMmQ0OTczMDEzNjY4NGE3OTg5NTZlYmUxNGRhNDMyYjYiLCJwIjoiaiJ9
## Things to check

### DIFF
--- a/app/models/hackney/court_outcome_helper.rb
+++ b/app/models/hackney/court_outcome_helper.rb
@@ -1,0 +1,21 @@
+module Hackney
+  module CourtOutcomeHelper
+    def human_readable_outcome(code)
+      code_mapping = {
+        'AGP' => 'Adjourned generally with permission to restore',
+        'AND' => 'Adjourned to next open date',
+        'AAH' => 'Adjourned to another hearing date',
+        'ADH' => 'Adjourned for directions hearing',
+        'ADT' => 'Adjourned on terms',
+        'OPF' => 'Outright possession forthwith',
+        'OPD' => 'Outright possession with date',
+        'SOT' => 'Suspension on terms',
+        'STO' => 'Struck out',
+        'WIT' => 'Withdrawn on the day',
+        'SOE' => 'Stay of execution'
+      }
+
+      code_mapping[code]
+    end
+  end
+end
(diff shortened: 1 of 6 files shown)


app/models/hackney/court_outcome_helper.rb
```diff
@@ -0,0 +1,21 @@
+module Hackney
+  module CourtOutcomeHelper
+    def human_readable_outcome(code)
+      code_mapping = {
+        'AGP' => 'Adjourned generally with permission to restore',
```
antonyoneill, line 5: I'm losing track of the name of the file, but can you use the static variables for this please? 

Might want to wait for #552 to get the court outcome codes in the new file

spec/lib/hackney/income/update_court_case_and_sync_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+        court_case_params: court_case_params
+      ).and_return(court_case)
+
+      expect(add_action_diary_and_sync_case_usecase).not_to receive(:execute)
+        .with(
+          username: username,
+          tenancy_ref: tenancy_ref,
+          action_code: 'IC6',
+          comment: 'Court outcome added: Struck out'
+        )
```
antonyoneill, line 72: Wouldn't bother passing the parameters, just check it's not called 

```suggestion
      expect(add_action_diary_and_sync_case_usecase).not_to receive(:execute)
```

antonyoneill (approved):
Looks good! just two minor things pls

app/models/hackney/court_outcome_helper.rb
antonyoneill, line 5:
```suggestion
        Hackney::Tenancy::CourtOutcomeCodes::ADJOURNED_GENERALLY_WITH_PERMISSION_TO_RESTORE => 'Adjourned generally with permission to restore',
```

antonyoneill (approved):

